PR #1, opened 2017-05-12
Initial

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1400

- [x] circle
- [x] README

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,15 @@
 # retagger
 
-`retagger` is a tool for managing retagging third party images
+`retagger` is a tool for managing retagging third party images.
+
+`retagger` is used to move all images required by Giant Swarm to our registry, while maintaining integrity of images - that is, we should be able to determine where all images came from. The alternative is manually retagging and pushing images, which means we lose a degree of accountability.
+
+A list of images (see `images.go`) is maintained, with a list of sha hashe and tag pairs. The `retagger` goes through each image, and then each sha hash and tag pair. The image is pulled by the sha hash, to overcome issues where a specific tag is rewritten, retagged with the tag, and pushed to our registry.
+
+The image name is also rewritten. For example, `quay.io/coreos/hyperkube` becomes `quay.io/giantswarm/hyperkube`, `prom/prometheus` becomes `quay.io/giantswarm/prometheus`.
+
+We attempt to pull the image from our registry first, to avoid unnecessary pulling of other images.
+
+We currently only support public images, both for pulling and pushing.
+
+The `retagger` works inside a CI build. On merges to master, the binary is executed. The workflow is to add a new image / sha tag pair in a PR, review it, and then merge. The `retagger` will take care that the image is handled. Operators still need to create repositories in the registry.
--- a/README.md
+++ b/README.md
@@ -1,10 +1,12 @@
 # retagger
 
+[![CircleCI](https://circleci.com/gh/giantswarm/retagger.svg?style=shield)](https://circleci.com/gh/giantswarm/retagger)
+
 `retagger` is a tool for managing retagging third party images.
 
 `retagger` is used to move all images required by Giant Swarm to our registry, while maintaining integrity of images - that is, we should be able to determine where all images came from. The alternative is manually retagging and pushing images, which means we lose a degree of accountability.
 
-A list of images (see `images.go`) is maintained, with a list of sha hashe and tag pairs. The `retagger` goes through each image, and then each sha hash and tag pair. The image is pulled by the sha hash, to overcome issues where a specific tag is rewritten, retagged with the tag, and pushed to our registry.
+A list of images (see `images.go`) is maintained, with a list of sha hashes and tag pairs. The `retagger` goes through each image, and then each sha hash and tag pair. The image is pulled by the sha hash, to overcome issues where a specific tag is rewritten, retagged with the tag, and pushed to our registry.
 
 The image name is also rewritten. For example, `quay.io/coreos/hyperkube` becomes `quay.io/giantswarm/hyperkube`, `prom/prometheus` becomes `quay.io/giantswarm/prometheus`.
 
@@ -12,4 +14,4 @@ We attempt to pull the image from our registry first, to avoid unnecessary pulli
 
 We currently only support public images, both for pulling and pushing.
 
-The `retagger` works inside a CI build. On merges to master, the binary is executed. The workflow is to add a new image / sha tag pair in a PR, review it, and then merge. The `retagger` will take care that the image is handled. Operators still need to create repositories in the registry.
+The `retagger` works inside a CI build. On merges to master, the binary is executed. The workflow is to add a new image / sha tag pair in a PR, review it, and then merge. The `retagger` will take care that the image is handled. Users will still need to create repositories in the registry.
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,20 @@
+machine:
+  services:
+    - docker
+
+dependencies:
+  override:
+    - |
+      wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
+    - chmod +x ./architect
+    - ./architect version
+
+test:
+  override:
+    - ./architect build
+
+deployment:
+  master:
+    branch: initial
+    commands:
+      - ./retagger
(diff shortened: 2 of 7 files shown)


circle.yml
```diff
@@ -0,0 +1,20 @@
...
+
+test:
+  override:
+    - ./architect build
+
+deployment:
+  master:
+    branch: initial
+    commands:
+      - ./retagger
```
JosephSalisbury: See here, this is where we do the actual retagging

images.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+type Tag struct {
+	Sha string
+	Tag string
+}
+
+var Images = []Image{
+	Image{
+		Name: "quay.io/coreos/hyperkube",
```
JosephSalisbury: Only adding hyperkube for the moment, I'll move everything we need once this PR is sorted

JosephSalisbury:
See https://circleci.com/gh/giantswarm/retagger/3 for example of the retagging in action, and https://quay.io/repository/giantswarm/hyperkube for the resulting repo

circle.yml
```diff
@@ -0,0 +1,20 @@
...
+    - chmod +x ./architect
+    - ./architect version
+
+test:
+  override:
+    - ./architect build
+
+deployment:
+  master:
+    branch: initial
```
JosephSalisbury: Note to self, change this to master before the merge

README.md
```diff
@@ -1,3 +1,15 @@
 # retagger
 
-`retagger` is a tool for managing retagging third party images
+`retagger` is a tool for managing retagging third party images.
+
+`retagger` is used to move all images required by Giant Swarm to our registry, while maintaining integrity of images - that is, we should be able to determine where all images came from. The alternative is manually retagging and pushing images, which means we lose a degree of accountability.
+
+A list of images (see `images.go`) is maintained, with a list of sha hashe and tag pairs. The `retagger` goes through each image, and then each sha hash and tag pair. The image is pulled by the sha hash, to overcome issues where a specific tag is rewritten, retagged with the tag, and pushed to our registry.
```
xh3b4sd: `hashe` -> `hash`

README.md
```diff
@@ -1,3 +1,15 @@
 # retagger
```
xh3b4sd: The circle and eventual quay badges would be awesome.

README.md
```diff
@@ -1,3 +1,15 @@
...
+
+A list of images (see `images.go`) is maintained, with a list of sha hashe and tag pairs. The `retagger` goes through each image, and then each sha hash and tag pair. The image is pulled by the sha hash, to overcome issues where a specific tag is rewritten, retagged with the tag, and pushed to our registry.
+
+The image name is also rewritten. For example, `quay.io/coreos/hyperkube` becomes `quay.io/giantswarm/hyperkube`, `prom/prometheus` becomes `quay.io/giantswarm/prometheus`.
+
+We attempt to pull the image from our registry first, to avoid unnecessary pulling of other images.
+
+We currently only support public images, both for pulling and pushing.
+
+The `retagger` works inside a CI build. On merges to master, the binary is executed. The workflow is to add a new image / sha tag pair in a PR, review it, and then merge. The `retagger` will take care that the image is handled. Operators still need to create repositories in the registry.
```
xh3b4sd: > Operators still need to create repositories in the registry. 

What does that mean?

images.go
```diff
@@ -0,0 +1,42 @@
...
+		Tags: []Tag{
+			Tag{
+				Sha: "5ff22b5c65d5b93aa948b79028dc136a22cda2f049283103f10bd45650b47312",
+				Tag: "v1.5.2_coreos.2",
+			},
+		},
+	},
+}
```
xh3b4sd: I am wondering if this is the right place for this configuration. I see this as configuration for a G8S installation like we already have in architect. What do you think in terms of future directions? Can/should we consolidate the image table with the rest of the settings a G8S installation needs? I can see this getting out of hand when it is in different places. How will the workflow look like when we have different installations using different registry endpoints? E.g. a customer probably wants to have its own registry and we need to sync our images with this one instead of quay. Thinking loud here.

images.go
```diff
@@ -0,0 +1,42 @@
...
+
+func RetaggedName(registry, organisation string, image Image) string {
+	parts := strings.Split(image.Name, "/")
+
+	return fmt.Sprintf("%v/%v/%v", registry, organisation, parts[len(parts)-1])
```
xh3b4sd: Would be cool to have constants for these formats and decent godoc.

kopiczko: or `path.Join`

main.go
```diff
@@ -0,0 +1,72 @@
+package main
+
+import (
+	"log"
+	"os"
+	"os/exec"
+)
+
+func main() {
+	registry := os.Getenv("REGISTRY")
```
kopiczko: `REGISTRY` is a bit generic. How about prefixing all env vars with `RETAGGER_`?

main.go
```diff
@@ -0,0 +1,72 @@
...
+	if registry == "" {
+		log.Fatalf("could not get registry\n")
+	}
+
+	registryOrganisation := os.Getenv("REGISTRY_ORGANISATION")
```
kopiczko: Maybe a struct instead of prefixing everything with `registry` or drop the prefix at all.

main.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	login := exec.Command("docker", "login", "--email=\" \"", "-u", registryUsername, "-p", registryPassword, registry)
+	if err := Run(login); err != nil {
+		log.Fatalf("could not login to registry: %v\n", err)
+	}
+
+	for _, image := range Images {
+		for _, tag := range image.Tags {
+			log.Printf("managing: %v, %v, %v\n", image.Name, tag.Sha, tag.Tag)
```
kopiczko: Newlines aren't necessary in log.

images.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func ImageWithTag(image, tag string) string {
```
kopiczko: I know functions are exported to avoid names collisions, but that allows unused code.

main.go
```diff
@@ -0,0 +1,72 @@
...
+			retaggedNameWithTag := ImageWithTag(retaggedName, tag.Tag)
+
+			log.Printf("checking if image has already been retagged")
+			pullRetag := exec.Command("docker", "pull", retaggedNameWithTag)
+			if err := Run(pullRetag); err == nil {
+				log.Printf("retagged image already exists, skipping\n")
+				continue
+			} else {
+				log.Printf("retagged image probably does not exist: %v\n", err)
+			}
```
kopiczko: It's sad there isn't a better way.
BTW can't we use docker lib?

images.go
kopiczko: Hmm, I would argue that godoc for a 2 loc function is not necessary. Also this format is simple. Does it make sense to create a constant for it and have to jump to see it's just `path.Join`?

images.go
JosephSalisbury: Eh, functions are exported because I didn't really think about it :D

JosephSalisbury: This is used in `main.go`, though

README.md
JosephSalisbury: I meant people. Changing to `Users still need ...`

README.md
JosephSalisbury: Should be `hashes`, I think

README.md
JosephSalisbury: Adding circle. No dockerfile for quay badge

images.go
kopiczko: I know, but if it was `imageWithTag` and it was unused compiler would shout. Nothing big to worry about.

images.go
JosephSalisbury: We talked conceptually. Retagger handles third party images moving to quay. We may need to think about how to synchronise images to local registries in the future.

main.go
JosephSalisbury: I didn't know this, actually

https://golang.org/pkg/log/#Output
A newline is appended if the last character of s is not already a newline.

That's really useful. I'll strip the newlines.

main.go
JosephSalisbury: No strong feeling. We don't tend to prepend the name to env vars for CI tools. Happy to change if a blocker.

images.go
JosephSalisbury: I'm not too fussed either way. Blocker?

JosephSalisbury:
@xh3b4sd @kopiczko mind having a round two?

main.go
kopiczko: I agree it's fine for CI.

PR #5, opened 2017-05-25
Retags golang 1.8.3

JosephSalisbury:


images_test.go
```diff
@@ -22,6 +22,12 @@ func TestRetaggedName(t *testing.T) {
 			},
 			expectedName: "quay.io/giantswarm/prometheus",
 		},
+		{
```
JosephSalisbury: This isn't needed for the retag, I just wanted to check that the retagger can correctly rename docker library images

PR #25, opened 2017-10-19
Add flannel

calvix:
add latest flannel image


jgsqware:
in term of go code LGTM

PR #29, opened 2017-11-22
Adds nginx and kube-state-metrics images

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1482


images.go
```diff
@@ -111,6 +124,15 @@ var Images = []Image{
 			},
 		},
 	},
+	Image{
```
JosephSalisbury: This is me alphabetising :D

PR #34, opened 2017-11-28
Revert "Add hyperkube 1.9 alpha"

r7vme:
This reverts commit b8f10a0db5b915eac9d1830d4e364c1034162913.

This image in hyperkube-git, so it fails. I'll just revert for now, will fix properly later.


calvix:
btw retagger does not delete  images so even if its removed it will be available in the quay repo

r7vme:
Image was not created for this. We are good imo.

PR #37, opened 2017-12-01
Retag hyperkube 1.9 beta for Azure operator

r7vme:


kopiczko:
JFYI azure-operator still uses TPR, so we can't just use 1.9. We are bound to 1.7. I can try to make that straight ASAP, but we are short on time.

PR #39, opened 2017-12-21
Retagge elasticsearch and kibana 6.1.1

jgsqware:

### DIFF
--- a/images.go
+++ b/images.go
@@ -17,6 +17,24 @@ type Tag struct {
 
 var Images = []Image{
 	Image{
+		Name: "docker.elastic.co/elasticsearch/elasticsearch",
+		Tags: []Tag{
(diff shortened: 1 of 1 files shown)


images.go
```diff
@@ -223,6 +223,24 @@ var Images = []Image{
...
+	},
+	Image{
+		Name: "docker.elastic.co/kibana/kibana",
+		Tags: []Tag{
+			Tag{
+				Sha: "6cccb1dc8b9caf656dd8c8527dcae5ff9d90e33f74f9d563f18f829b96fbbbff",
+				Tag: "6.1.1",
+			},
+		},
+	},
```
kopiczko: I think we are trying to have an alphabetical order of images here. Please fix that and go.

PR #40, opened 2017-12-21
add-fluentd

jgsqware:


kopiczko:
Note that when CI is broken this serves no purpose :(

https://gigantic.slack.com/archives/C03CPNRTS/p1513852102000456

PR #44, opened 2018-01-23
Add nginx-ingress-controller-0.10.0

corest:


images.go
```diff
@@ -107,6 +107,15 @@ var Images = []Image{
 			},
 		},
 	},
+	Image{
```
JosephSalisbury: Please alphabetise this list :D

PR #74, opened 2018-04-13
replace hyperkube 1.10.0 -> 1.10.1

kopiczko:


JosephSalisbury:
I would have added the new tag, instead of replacing the older one.

kopiczko:
Older one is not used anywhere. Should it be left anyways?

JosephSalisbury:
I would prefer it, but it's not a big issue

PR #88, opened 2018-05-22
Add images needed for GS services

puja108:
towards https://github.com/giantswarm/adidas/issues/320

already created busybox repo in quay

I added latest versions of some images, so we can use this chance to also upgrade to latest


marians:
Is it necessary to keep the old versions around in the struct? If retagger's only task is to pull from source and push to quay, doing this once should be enough. So we don't need to grow the images struct, correct?

puja108:
@marians very good point, I'm just not 100% sure about removing, will start a thread in sig-releng

PR #93, opened 2018-06-07
Adds hyperkube 1.10.4 image

rossf7:


images.go
```diff
@@ -227,6 +227,10 @@ var Images = []Image{
 				Sha: "00d814b1f7763f4ab5be80c58e98140dfc69df107f253d7fdd714b30a714260a",
 				Tag: "v1.10.3",
 			},
+			Tag{
+				Sha: "2b44d8a1bdb323a56f08ed803eb6cd4f9b7faef5d407c9ebd14d509b8d2f5276",
+				Tag: "v1.10.4",
+			},
```
rossf7: I've added v1.10.4 here but not sure if I should remove v1.10.3 and v1.10.2? Once the image has been retagged do we still need it here?

tuommaki: In general I like the idea of removing old images, but I'd probably test this separately from here.

rossf7: Good call👍 I'll merge this now and do a follow up PR for removing the old images.

PR #103, opened 2018-06-29
Use v2 registry API

fgimenez:
Uses registry v2 API for getting existing image tags on the target registry. Includes code for requesting a token according to https://docs.docker.com/registry/spec/auth/token/. Tested with Quay and aliyun.

### DIFF
--- a/registry.go
+++ b/registry.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
 	"log"
 	"net/http"
 	"os/exec"
+	"strings"
 
 	"github.com/giantswarm/microerror"
 )
@@ -27,6 +29,14 @@ type Registry struct {
 	password     string
 	token        string
 	username     string
+}
+
+type TagsListResponse struct {
+	Tags []string `json:"tags"`
+}
+
+type TokenRequestResponse struct {
+	Token string `json:"token"`
 }
 
 func NewRegistry(cfg *RegistryConfig) (*Registry, error) {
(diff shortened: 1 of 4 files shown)


registry.go
```diff
@@ -75,12 +85,18 @@ func (r *Registry) Login() error {
 }
 
 func (r *Registry) CheckImageTagExists(image, tag string) (bool, error) {
-	url := fmt.Sprintf("https://%s/api/v1/repository/%s/tag/%s/images", r.host, ImageName(r.organisation, image), tag)
+	url := fmt.Sprintf("https://%s/v2/%s/tags/list", r.host, ImageName(r.organisation, image))
```
rossf7: <3

registry.go
```diff
@@ -108,3 +134,80 @@ func (r *Registry) Retag(image, sha, tag string) (string, error) {
 	}
 	return retaggedNameWithTag, nil
 }
+
+func (r *Registry) getToken(req *http.Request) (string, error) {
+	const authorizationHeaderKey = "Www-Authenticate"
```
rossf7: Does the casing here matter? I did a quick google and saw some examples also using `WWW-Authenticate`.

registry_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+func Test_GetAuthURL(t *testing.T) {
+	tcs := []struct {
+		challenge     string
+		expectedURL   string
+		expectedError bool
+		description   string
+	}{
+		{
+			description: "base case",
```
rossf7: Nitpick but could we use the "case 0:" style we have in `fmt`? 

I find it quicker to track down test failures.

registry.go
fgimenez: Very good point, on aliyun we have `Www-Authenticate` but even in the RFC they talk about `WWW-Authenticate` https://tools.ietf.org/html/rfc6750#section-3. I've changed the code to iterate through the header keys and check against a lowercased version, PTAL.

registry_test.go
fgimenez: Sure, done.

registry.go
rossf7: Nice! Good that the comparison is now case insensitive. I can definitely imagine some other registry using a different case and that could be nasty to debug!

PR #112, opened 2018-07-11
Add k8s-setup-network-environment

fgimenez:
Adds `k8s-setup-network-environment` image used in node initialization. This is our own image, adding it here mainly for syncing in aliyun registry.

### DIFF
--- a/images.go
+++ b/images.go
@@ -394,6 +394,15 @@ var Images = []Image{
 			},
 		},
 	},
+	Image{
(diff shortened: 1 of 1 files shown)


fgimenez:
> Still feels weird to add our own images, but guess thats the way to go atm

+1, quay is still the central repository for the build system. retagger is the mean we have to distribute images, even our own.

PR #120, opened 2018-07-31
Orders images alphabetically, adds test

JosephSalisbury:
The images list was a mess, so I've added a test that it is alphabetical by image name (so we have to keep it nice and ordered), and then also ordered it alphabetically

### DIFF
--- a/images.go
+++ b/images.go
@@ -17,95 +17,29 @@ type Tag struct {
 
 var Images = []Image{
 	Image{
-		Name: "coredns/coredns",
-		Tags: []Tag{
-			Tag{
-				Sha: "399cc5b2e2f0d599ef22f43aab52492e88b4f0fd69da9b10545e95a4253c86ce",
-				Tag: "1.1.1",
-			},
-		},
-	},
-	Image{
-		Name: "quay.io/calico/node",
-		Tags: []Tag{
-			Tag{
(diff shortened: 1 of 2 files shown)


images_test.go
```diff
@@ -93,3 +96,18 @@ func TestShaName(t *testing.T) {
 		}
 	}
 }
+
+func TestImagesOrderedAlphabetically(t *testing.T) {
```
xh3b4sd: You can imagine how much I like that. xD

JosephSalisbury: This PR is a love letter to you <3

images.go
```diff
@@ -179,6 +102,17 @@ var Images = []Image{
 			},
 		},
 	},
+	Image{
+		// see https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-ds.yaml
+		Name: "gcr.io/google_containers/fluentd-elasticsearch",
```
JosephSalisbury: For the record, I changed this from `gcr.io/google-containers/fluentd-elasticsearch` to `gcr.io/google_containers/fluentd-elasticsearch` - note the change from a dash to an underscore.

It took me ages to notice, and it changes the alphabetisation (and both urls work)

PR #126, opened 2018-08-16
Retagg oauth2_proxy

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/3958

Use community image that supports token renewal with OICD.

See https://github.com/bitly/oauth2_proxy/pull/621

### DIFF
--- a/images.go
+++ b/images.go
@@ -177,6 +177,15 @@ var Images = []Image{
 		},
 	},
 	Image{
+		Name: "gfkse/oauth2_proxy",
+		Tags: []Tag{
+			Tag{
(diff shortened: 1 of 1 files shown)


fgimenez:
We will need to create the image in aliyun registry. @puja108 can you create an user for @indyfree ?

indyfree:
To understand correctly @fgimenez. We have the Cirlce aliyun retag job in place but need to create the image in aliyun manually first?

https://github.com/giantswarm/retagger/blob/bd22e7ff4dd6b53db7ff39d461193eb8906aa313/.circleci/config.yml#L31-L33

fgimenez:
@indyfree yes, it is the same as with quay, we have a job for each registry and we need to create the repository before pushing images.

PR #144, opened 2018-10-24
update golang to v1.11.1

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/4461

### DIFF
--- a/images.go
+++ b/images.go
@@ -202,7 +202,7 @@ var Images = []Image{
 		Name: "golang",
 		Tags: []Tag{
 			Tag{
-				Sha: "a325153f1f1e2edb76a76ad789aff172b89dd6178e8f74d39ef87c04d87d1961",
+				Sha: "63ec0e29aeba39c0fe2fc6551c9ca7fa16ddf95394d77ccee75bc7062526a96c",
 				Tag: "1.11.1",
 			},
 		},
--- a/images.go
+++ b/images.go
@@ -202,8 +202,8 @@ var Images = []Image{
 		Name: "golang",
 		Tags: []Tag{
 			Tag{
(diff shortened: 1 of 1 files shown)


images.go
```diff
@@ -202,8 +202,8 @@ var Images = []Image{
 		Name: "golang",
 		Tags: []Tag{
 			Tag{
-				Sha: "a325153f1f1e2edb76a76ad789aff172b89dd6178e8f74d39ef87c04d87d1961",
-				Tag: "1.10.4",
+				Sha: "63ec0e29aeba39c0fe2fc6551c9ca7fa16ddf95394d77ccee75bc7062526a96c",
```
corest: Aren't we just adding new tag usually? So that if there are references somewhere - we will not break it. Or is that a purpose - break all the builds and force them to update tag?

TheoBrigitte: I was not sure how this work, I though we needed to add the tag and sha from the target image according to readme: https://github.com/giantswarm/retagger#tags.

TheoBrigitte: reverted sha, PTAL

images.go
```diff
@@ -203,7 +203,7 @@ var Images = []Image{
 		Tags: []Tag{
 			Tag{
 				Sha: "a325153f1f1e2edb76a76ad789aff172b89dd6178e8f74d39ef87c04d87d1961",
```
MarcelMue: You were right the first time, we switch both the `sha` and `tag` here usually:
https://github.com/giantswarm/retagger/pull/138/files

TheoBrigitte: Fine, I'll revert the revert. This is a bit comfusing

PR #148, opened 2018-11-01
Maintain image configuration in YAML file

marians:
YAML should be easier to maintain for humans than a Go file. It also opens this data up for handling in other tools and automation, and allows for separation of code and config.

The YAML data is validated in two steps: A first CI step runs `yamllint`, and then of course the parsing will fail at runtime.

Technically this moves the validation away from compilation to CI and execution. However, in practice, many retagger PRs were simply made by editing the `images.go` file via Github.com and leaving compilation to CI. This should become easier by using YAML instead of Go code.

### YAML Schema

Example entry:

```yaml
- name: gcr.io/google_containers/fluentd-elasticsearch
  comment: see https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-elasticsearch/fluentd-es-ds.yaml
  tags:
  - sha: b8c94527b489fb61d3d81ce5ad7f3ddbb7be71e9620a3a36e2bede2f2e487d73
    tag: v2.0.4
```

The YAML schema is derived almost exactly from the Golang structs in use.

One addition: The optional `comment` field exists to formally add a comment, e. g. regarding provenance of the image.


marians:
I did some more changes.

- README completely updated
- Changed order in YAML: SHA first, then tag. This should reflect the logical order.

and more things.

PR #171, opened 2018-12-28
Update coredns, calico and etcd

r7vme:
_Retagger_

_!! WARNING !!_
_If you are adding new image to retagger be sure to create repository in both container registry !!!!_

---

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -8,8 +8,8 @@
     tag: 1.28.3
 - name: coredns/coredns
   tags:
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -94,10 +94,6 @@
   tags:
   - sha: b9a31857e86e9cf43552605bd7f3c990c123f8792ab6bea8f499db1a1bdb7d53
     tag: 5.4.2
-- name: jetstack/kube-lego
```
r7vme: kube-lego deprecated long time ago.

PR #193, opened 2019-02-19
All overriding repository name

jgsqware:

Regarding Retagger again: We have the case that we want to have `bitnami/redis` images in our repo. But we already have the `redis` repository mirroring.

that means `giantswarm/redis` already exists.

This PR add a new field `overrideName` to have the possibility to specify the destination repository name like this:

```yaml
- name: redis
  tags:
  - sha: 002a1870fa2ffd11dbd7438527a2c17f794f6962f5d3a4f048f848963ab954a8
    tag: 3.2.11-alpine
  - sha: 87275ecd3017cdacd3e93eaf07e26f4a91d7f4d7c311b2305fccb50ec3a1a8cd
    tag: 4.0.9
- name: bitnami/redis
  overrideName: bitnami-redis
  tags:
  - sha: 87275ecd3017cdacd3e93eaf07e26f4a91d7f4d7c311b2305fccb50ec3a1a8cd
    tag: 4.0.9```

### DIFF
--- a/images.go
+++ b/images.go
@@ -11,9 +11,10 @@ import (
 
 // Image defines the data we process about a docker image.
 type Image struct {
-	Name    string `yaml:"name"`
-	Comment string `yaml:"comment,omitempty"`
-	Tags    []Tag  `yaml:"tags"`
+	Name         string `yaml:"name"`
+	Comment      string `yaml:"comment,omitempty"`
+	OverrideName string `yaml:"overrideName,omitempty"`
+	Tags         []Tag  `yaml:"tags"`
 }
 
 // Tag represents a specific version of a docker image, represented by a tag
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: images.go [hunk 1]
```diff
@@ -11,9 +11,10 @@ import (
 
 // Image defines the data we process about a docker image.
 type Image struct {
-	Name    string `yaml:"name"`
-	Comment string `yaml:"comment,omitempty"`
-	Tags    []Tag  `yaml:"tags"`
+	Name         string `yaml:"name"`
+	Comment      string `yaml:"comment,omitempty"`
+	OverrideName string `yaml:"overrideName,omitempty"`
```
JosephSalisbury: Would `OverrideRepoName` make more sense? I feel like `Name` and `OverrideName` is a bit confusing

jgsqware: LGTM

PR #197, opened 2019-02-19
Add quay and aliyun creator

jgsqware:

Add the creation of Quay and Aliyun repository and notifications if needed.


aliyun/cr/api.go
```diff
@@ -0,0 +1,101 @@
+package cr
```
JosephSalisbury: cr?

jgsqware: aliyun refer Container Registry as `cr` that's why

you prefer `containerregistry` ?

JosephSalisbury: I think idiomatic golang would be `registry`. Single words for package names and all that.

tfussell:
I think this is no longer necessary. I'm going to close it. Let me know if you disagree @jgsqware.

PR #217, opened 2019-05-14
Add custom rebuild logic

corest:
Add logic for custom images rebuilds.
---
```
- name: coredns/coredns                                                         
  tags:                                                                         
  - sha: e83beb5e43f8513fa735e77ffc5859640baea30a882a11cc75c4c3244a737d3c       
    tag: 1.5.0                                                                  
    dockerfileOptions:                                                          
      - EXPOSE 1053
```

This will retag `coredns/coredns:1.5.0` into `quay.io/giantswarm/coredns:1.5.0` with th and additionally rebuild `quay.io/giantswarm/coredns:1.5.0-giantswarm` from Dockerfile
```
FROM coredns/coredns;1.5.0
EXPOSE 1053
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -24,3 +24,10 @@ var invalidAuthenticateChallengeError = microerror.New("invalid authenticate cha
 func IsInvalidAuthenticateChallenge(err error) bool {
 	return microerror.Cause(err) == invalidAuthenticateChallengeError
 }
+
+var invalidTemplateError = microerror.New("invalidTemplateError")
+
+// IsInvalidTemplate asserts invalidTemplateError.
+func IsInvalidTemplate(err error) bool {
+	return microerror.Cause(err) == invalidTemplateError
+}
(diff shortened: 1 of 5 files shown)


corest:
Another use-case - having base image with non-root:
```
- name: alpine
  tags:
  - sha: 7df6db5aa61ae9480f52f0b3a06a140ab98d427f86d8d5de0bedab9b8df6b1c0
    tag: "3.9"
    dockerfileOptions:
      - USER giantswarm
```

kopiczko:
What if I'd like to have two custom images based on alpine:3.9?

corest:
There is no such option. Can you describe real-world case?

Main idea here is ability to make upstream containers we are using, ready to run as non root.

kopiczko:
> Can you describe real-world case?

Yes for circleci docker builds I'd like to have images with different tools installed based on alpine. E.g. with `go` and `jq`. Then you also want to have your non-root image.

corest:
@kopiczko added support for multiple custom images

PR #218, opened 2019-05-15
Custom alpine coredns

corest:
This will come after https://github.com/giantswarm/retagger/pull/217 merged


corest:
@giantswarm/sig-releng ptal pls and sorry for ping but that blocks me from moving into psps for tenant cluster components

PR #248, opened 2019-08-20
Retag kong and kong-ingress-controller

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/6712

TODO: I still need to create Aliyun images. 2FA is setup on my old phone which is at home. 🤦‍♂ 

I'll fix before merging.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -197,6 +197,14 @@
   tags:
   - sha: 7d4d712a2686da99d37580b4e2f45eb658b74e4b01caf67c1099adc294b96b52
     tag: v0.6.0
+- name: kong
+  tags:
+  - sha: df7bb41be97ec490d08e7fb5acb2de626042a1eedb16405bc8504814a67140c0
+    tag: 1.2.2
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -197,6 +197,14 @@
   tags:
   - sha: 7d4d712a2686da99d37580b4e2f45eb658b74e4b01caf67c1099adc294b96b52
     tag: v0.6.0
+- name: kong
+  tags:
+  - sha: df7bb41be97ec490d08e7fb5acb2de626042a1eedb16405bc8504814a67140c0
+    tag: 1.2.2
+- name: kong-docker-kubernetes-ingress-controller.bintray.io/kong-ingress-controller
```
rossf7: This is a public jFrog registry. I'm not sure why it's not on Docker Hub.

rossf7: OK it's because the enterprise version is there.

kong-docker-kong-enterprise-edition-docker.bintray.io/kong-enterprise-edition

rossf7:
Aliyun sorted.

PR #250, opened 2019-08-21
Add latest busybox image

rossf7:
Kong uses busybox:latest as a wait image and ours is quite old. Updating to use 1.31.0 which is latest.


rossf7:
Going with a ping.

PR #252, opened 2019-08-28
add ability to use different retagger configuration files

schjan:
Needed for easier testing of subsequent changes. :building_construction: 

_Towards https://github.com/giantswarm/giantswarm/issues/5968_


images-test.yaml
```diff
@@ -0,0 +1,11 @@
...
+  tags:
+    - sha: 769fddc7cc2f0a1c35abb2f91432e8beecf83916c421420e6a6da9f8975464b6
+      tag: "3.9"
+      customImages:
+        - tagSuffix: giantswarm
+          dockerfileOptions:
+            - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm
+            - USER giantswarm
+    - sha: 7df6db5aa61ae9480f52f0b3a06a140ab98d427f86d8d5de0bedab9b8df6b1c0
+      tag: "3.7"
```
MarcelMue: Why no new line add the end of the file?

main.go
```diff
@@ -1,14 +1,28 @@
...
 import (
+	"flag"
 	"fmt"
 	"log"
 	"net/http"
 	"os"
 	"os/exec"
 )
 
+var configFile = flag.String("f", "images.yaml", "retagger config (images.yaml) file to use.")
```
MarcelMue: I would prefer to use cobra/viper here. It would be more consistent with other projects then.

And then we might have an easier time by just switching to the structure of other commandline tools - e.g. `devctl`
https://github.com/giantswarm/devctl

registry.go
```diff
@@ -116,17 +131,13 @@ func (r *Registry) CheckImageTagExists(image, tag string) (bool, error) {
...
-			}
+			return nil, microerror.Mask(err)
 		}
-		return false, nil
+
+		return tagResponse.Tags, nil
 	default:
 		log.Printf("could not check retag status: %v", res.StatusCode)
-		return false, nil
+		return nil, nil
```
MarcelMue: Why does this not return some error?

schjan:
Thx @MarcelMue for the comments.

I wanted to add that `-f` flag as simple as possible to be able to test the current state of the application before doing more complex changes.
Also my next PR would be to create a simple e2e test which needs this.

After that I want to start to restructure the code to better align to the `devctl` standards i.e. using cobra and then implement the new features.

MarcelMue:
Let's increase the review circle here. I am not opinionated about how to progress here, so let's check if others are okay with this.

PR #254, opened 2019-08-29
Create a simple e2e test

schjan:
Also remove custom Registry Client as it is buggy when it comes to authentication.


Gopkg.toml
```diff
@@ -36,3 +36,7 @@
 [[constraint]]
   branch = "master"
   name = "github.com/giantswarm/backoff"
+
+[[constraint]]
+  branch = "master"
+  name = "github.com/nokia/docker-registry-client"
```
xh3b4sd: Why don't we use `github.com/docker/docker`?

schjan: docker/docker is a client for your local docker instance.
Sadly there is no offical API client for Docker Registries.

schjan:
What do you think shoud I create a aliyun e2e test in this PR, too?

Gopkg.toml
xh3b4sd: What is with https://godoc.org/github.com/docker/distribution/registry/client?

schjan: I don't see how this implements our needed functionality.

PR #255, opened 2019-08-29
Refactor repository to match devctl structure

schjan:
... or at least _any_ structure :smiley: 

The PR for now is mostly copy pasting stuff from old package _main_ to separate packages and get everything working again.
Didn't really know how to make this PR smaller so I created the PR in this state. After that I will create additional smaller PRs to fix bugs etc.


pkg/config/images_test.go
```diff
@@ -1,10 +1,35 @@
...
+		organization   string
+		image          string
+		expectedResult string
+	}{
+		{
+			host:           "quay.io",
+			organization:   "acme",
+			image:          "gcr.io/foo/bar",
+			expectedResult: "quay.io/acme/bar",
+		},
```
MarcelMue: For later: Add a few more test cases.

pkg/config/config.go
```diff
@@ -0,0 +1,30 @@
...
+}
+
+func FromFile(filePath string) (*Config, error) {
+	yamlFile, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return nil, microerror.Maskf(err, "could not read file %s: #%v ", filePath)
+	}
+
+	var images []Image
+	err = yaml.Unmarshal(yamlFile, &images)
```
MarcelMue: Add a block here.

internal/env.go
```diff
@@ -0,0 +1 @@
+package internal
```
MarcelMue: Huh?

.gitignore
```diff
@@ -1,2 +1 @@
-retagger
-
+/retagger
```
MarcelMue: Add ignore for e2e test binary.

cmd/version/flag.go
```diff
@@ -0,0 +1,13 @@
+package version
```
MarcelMue: Not sure if we need this. More general I guess.

e2e/e2e_test.go
```diff
@@ -1,19 +1,21 @@
 // +build e2e
 
-package main
+package e2e
```
xh3b4sd: All other projects I know of have an `integration/test` directory. We should align this here. Feel free to do this in a separate PR.

main.go
```diff
@@ -1,103 +1,56 @@
...
-	}
+	"github.com/giantswarm/retagger/cmd"
+)
 
-	err := InitImages(*configFile)
+func main() {
+	err := mainE(context.Background())
 	if err != nil {
-		log.Fatalf("could not read retagger config %v", err)
+		panic(fmt.Sprintf("%#v\n", err))
```
xh3b4sd:
```suggestion
		panic(microerror.Stack(err))
```

pkg/config/error.go
```diff
@@ -0,0 +1 @@
+package config
```
xh3b4sd: Why do we need this empty file?

pkg/registry/error.go
```diff
@@ -0,0 +1,31 @@
+package registry
+
+import "github.com/giantswarm/microerror"
+
+var invalidConfigError = microerror.New("invalid config")
```
xh3b4sd: Why does this use `microerror.New` and the others define structs inline?

cmd/runner.go
```diff
@@ -0,0 +1,142 @@
...
+			if image.OverrideRepoName != "" {
+				r.logger.LogCtx(ctx, "message", "level", "debug", fmt.Sprintf("Override Name specified. Using %s as mirrored image name", image.OverrideRepoName))
+				imageName = image.OverrideRepoName
+			}
+			r.logger.LogCtx(ctx, "message", "level", "debug", fmt.Sprintf("managing: %v, %v, %v", imageName, tag.Sha, tag.Tag))
+
+			for _, customImage := range tag.CustomImages {
+				ok, err := destRegistry.CheckImageTagExists(imageName, tag.Tag)
+				if ok {
+					r.logger.LogCtx(ctx, "message", "level", "debug", fmt.Sprintf("rebuilt image %q with tag %q already exists, skipping", imageName, fmt.Sprintf("%s-%s", tag.Tag, customImage.TagSuffix)))
```
tfussell: Five levels of indentation is reaching the point of being difficult to read. I might break this inner loop into a separate function if it doesn't cause too much trouble.

schjan: I want to move the code starting somewhere around `for _, image := range conf.Images {` into a different package with the next PR. Didn't do it now as this would bloat up this PR even more.

tfussell: Fine with me. Keep PRs small :)

PR #257, opened 2019-09-06
create pkg/retagger for core logic

schjan:
_Towards https://github.com/giantswarm/giantswarm/issues/5968_


schjan:
Rebuilding images still doesn't use the golang `docker/client` as dealing with the buildcontext felt super awkward.

pkg/registry/registry_docker.go
```diff
@@ -0,0 +1,151 @@
...
+
+	authConfig := types.AuthConfig{
+		Username:      r.username,
+		Password:      r.password,
+		ServerAddress: r.host,
+	}
+	res, err := r.docker.RegistryLogin(context.Background(), authConfig)
+	if err != nil {
+		return microerror.Maskf(err, "could not login to registry")
+	} else if res.Status != "Login Succeeded" {
```
MarcelMue: The error matcher should check this probably.

.gitignore
```diff
@@ -1,2 +1,3 @@
 /retagger
 /e2e.test
+Dockerfile*
```
MarcelMue: New line.

pkg/registry/registry_docker.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	r.registryAuth = res.IdentityToken
+
+	return nil
+}
+
+func (r *Registry) PullImage(image string, sha string) error {
+	if sha == "" {
+		return microerror.Maskf(invalidArgumentError, "%s SHA should not be empty", image)
```
MarcelMue: Why is there a need to check the sha but not the image?

pkg/registry/registry_docker.go
```diff
@@ -0,0 +1,151 @@
...
+	retaggedNameWithTag := fmt.Sprintf("%s:%s", destinationImage, destinationTag)
+
+	dockerfile := Dockerfile{
+		BaseImage:         sourceImage,
+		Sha:               sha,
+		DockerfileOptions: dockerfileOptions,
+		Tag:               destinationTag,
+	}
+
+	f, err := os.Create(fmt.Sprintf("Dockerfile-%s", destinationTag))
```
MarcelMue: Not hyper important but having a `key.go` with stuff like `fmt.Sprintf("Dockerfile-%s", destinationTag)` as its own little function is pretty handy. 
We do this in operators as these functions can then be reused (see line 120)

pkg/retagger/executor.go
```diff
@@ -0,0 +1,79 @@
+package retagger
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+)
+
+func (r *Retagger) ExecuteJobs() error {
+	r.logger.Log("level", "debug", "message", fmt.Sprintf("start executing %d jobs", len(r.jobs)))
```
MarcelMue: There should then also be a message about it being done.

pkg/retagger/job.go
```diff
@@ -0,0 +1,94 @@
...
+type JobOptions struct {
+	// DockerfileOptions - list of strings we add for Dockerfile to build custom image.
+	DockerfileOptions []string
+
+	TagSuffix string
+
+	OverrideRepoName string
+}
+
+func FromConfig(images images.Images) ([]Job, error) {
```
MarcelMue: Why not `FromImages`?

pkg/registry/registry_docker.go
schjan: because legacy :)

pkg/images/images.go
```diff
@@ -31,25 +33,22 @@ type CustomImage struct {
 	DockerfileOptions []string `yaml:"dockerfileOptions"`
 }
 
-// Images is the data structure that will hold all image definitions.
-var Images []Image
-
 func ImageName(organisation string, image string) string {
```
MarcelMue: Not really part about this PR but still something:
IMO this should be called `Name` - the package is already called `Image` - `images.ImageName` stutters.

pkg/registry/registry_docker.go
```diff
@@ -0,0 +1,185 @@
...
+	err = r.logDocker(res)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (r *Registry) TagSha(sourceImage, sha, destinationImage, destinationTag string) (string, error) {
+	if sourceImage == "" {
```
MarcelMue: I thought about this and might have lead you in the wrong direction here:
Does it really make sense to check for empty values here? Is this something we should be doing?
I am not against this in this case here, just wondering as this is inconsistent across packages here now.

Not blocking.

schjan: Had the option to just delete the checking or extend it to all functions, so I decided the second

schjan: I think its generally better than getting awkward errors from docker/client or docker cli because some fields are empty and _things happen_:tm:

PR #266, opened 2019-10-02
Retag k8s 1.15.4, etcd 3.3.15, calico 3.9.1, remove old versions

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/6431


tfussell:
I feel like we should either keep all versions or only one version, but I see PRs that use both approaches. I would vote for keeping only the most recent because it keeps the images file cleaner.

PR #287, opened 2019-10-29
create image repository if not exists

calvix:
Create docker registry repository if its not created.

We simply ignore 404 repositories not know to the registry as push into such repos will create the repository for us (proper settings in quay and aliyun are set). So we treat non existing repository as if the repository was empty without any tags.

### DIFF
--- a/pkg/registry/registry.go
+++ b/pkg/registry/registry.go
@@ -111,7 +111,10 @@ func (r *Registry) ListImageTags(image string) ([]string, error) {
 	var tags []string
 	o := func() error {
 		imageTags, err := r.registryClient.Tags(images.Name(r.organisation, image))
-		if err != nil {
+		if IsRepositoryNotFound(err) {
+			r.logger.Log("level", "warning", "message", fmt.Sprintf("repository %s was not found in registry, retagger will try create the repository", image))
+			return nil
+		} else if err != nil {
 			return microerror.Mask(err)
 		}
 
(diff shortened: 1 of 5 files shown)


calvix:
tested with few use-cases and this seems to work now ( with settings for auto-creation for repository)

pkg/registry/registry.go
```diff
@@ -111,7 +111,10 @@ func (r *Registry) ListImageTags(image string) ([]string, error) {
 	var tags []string
 	o := func() error {
 		imageTags, err := r.registryClient.Tags(images.Name(r.organisation, image))
-		if err != nil {
+		if IsRepositoryNotFound(err) {
```
MarcelMue: Very cool if that works - can you share which settings you set to make this work? In quay for example?

MarcelMue: There was also a public user who asked for support of this behavior. 
https://github.com/giantswarm/retagger/issues/263

calvix: in aliyun, there is specific settings on resource `namespace`  (similar to organisation in quay` that is called `auto-create repository` so i checked that out

in quay you dont need extra box checked but the account need to have admin privileges so it can create repositories

That is basically all that needs to be done. It just needed few tries to get it straight.

calvix: I am not sure if it makes sense to document this somewhere since this seem very provider specific and any new registyr will probabbly have different settings

MarcelMue: Answer is totally sufficient to me. Makes sense :+1:

Thanks for taking care of this :tada:

PR #292, opened 2019-11-18
Add hyperkube 1.16.3 image

stone-z:
_Retagger_

---


stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/7428

PR #298, opened 2019-12-11
Retag new images automatically

stone-z:
This feature introduces lots more complexity.

The new behavior:
 - keeps the previous SHA/Tag behavior identical
 - introduces the `patterns` key
 - each `pattern` will be used to compile a regular expression
 - retagger pulls the tag list of each upstream repo and selects tags matching the pattern
 - each `pattern` may also specify whether to update the quay image if an upstream image changes
 - introduces the --whatif flag to print jobs but not run them

Towards https://github.com/giantswarm/giantswarm/issues/5968


stone-z:
Still todo:
 - allow 'real' regex in the pattern (currently matches literal special characters but we will need ^ to avoid pulling alpha and beta images, for example)
 - handle issue with GCR where it requires either an access token from the repo (e.g. hyperkube) or a workaround in order to list tags  -- this is in progress
 - updated docs and tests

cmd/flag.go
```diff
@@ -19,6 +20,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&f.Organization, "organization", "o", "giantswarm", "organization to tag images for")
 	cmd.Flags().StringVarP(&f.Username, "username", "u", "", "username to authenticate against registry")
 	cmd.Flags().StringVarP(&f.Password, "password", "p", "", "password to authenticate against registry")
+	cmd.Flags().BoolVar(&f.WhatIf, "whatif", false, "if set, will list jobs but not run them")
```
MarcelMue: We usually call this `dry-run`

pkg/images/images.go
```diff
@@ -9,10 +9,20 @@ type Images []Image
...
 type Image struct {
-	Name             string `yaml:"name"`
-	Comment          string `yaml:"comment,omitempty"`
-	OverrideRepoName string `yaml:"overrideRepoName,omitempty"`
-	Tags             []Tag  `yaml:"tags"`
+	Name             string       `yaml:"name"`
+	Comment          string       `yaml:"comment,omitempty"`
+	OverrideRepoName string       `yaml:"overrideRepoName,omitempty"`
+	Tags             []Tag        `yaml:"tags"`
+	Patterns         []TagPattern `yaml:"patterns"`
```
MarcelMue: Please keep this alphabetical only with `Name` as the first field.

The logic was that `Name` makes sense as a first field like in k8s and all following should be alphabetical. Unless you have some other reasoning for this order.

pkg/images/images.go
```diff
@@ -9,10 +9,20 @@ type Images []Image
...
+}
+
+// TagPattern represents a group of tags defined by a regular expression,
+// and stores some configuration for how to handle this group.
+type TagPattern struct {
```
MarcelMue: Same about alphabetical order here.

pkg/registry/registry.go
```diff
@@ -156,3 +159,47 @@ func (r *Registry) DeleteImage(image string, tag string) error {
 func (r *Registry) RetaggedName(image string) string {
 	return images.RetaggedName(r.host, r.organisation, image)
 }
+
+// GuessRegistryPath examines the given image string,
+// determines whether it describes a full image path,
+// is hosted on Docker hub, or belongs to the Docker standard library, and returns a URL representing the full path.
```
MarcelMue: Your godoc character limit seems to inconsistent :P

pkg/registry/registry.go
```diff
@@ -156,3 +159,47 @@ func (r *Registry) DeleteImage(image string, tag string) error {
...
+func (r *Registry) GetRepositoryFromPathString(path string) (string, error) {
+	url, err := r.GuessRegistryPath(path)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+	return r.GetRepositoryFromPath(url)
+}
+
+// GetRepositoryFromPath extracts the organization/image segment of a full image path.
+func (r *Registry) GetRepositoryFromPath(path *nurl.URL) (string, error) {
```
MarcelMue:
```suggestion
func (r *Registry) GetRepositoryFromPath(path *nurl.URL) string {
```

No reason to have error in the returns if there can't be an error.

pkg/registry/registry.go
```diff
@@ -2,8 +2,11 @@ package registry
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
+	nurl "net/url"
```
MarcelMue: Why is this import in its own block? Should be in the first block IMO.

pkg/registry/registry.go
```diff
@@ -156,3 +159,47 @@ func (r *Registry) DeleteImage(image string, tag string) error {
...
+
+	// Image string already has FQDN
+	if url.Hostname() != "" {
+		return nurl.Parse(image)
+	}
+
+	pathParts := strings.Split(strings.Trim(url.Path, "/"), "/")
+	if len(pathParts) == 2 {
+		// Image string is in the form organization/image, so we assume it is on docker hub
+		return nurl.Parse(fmt.Sprintf("https://registry.hub.docker.com/%s", image))
```
MarcelMue: This seems like a weird assumption to me. Do we make this assumption somewhere already?

pkg/registry/registry.go
```diff
@@ -156,3 +159,47 @@ func (r *Registry) DeleteImage(image string, tag string) error {
...
+	}
+
+	if len(pathParts) == 1 {
+		// Image is a single token, so must be part of the official Docker library
+		return nurl.Parse(fmt.Sprintf("https://registry.hub.docker.com/%s/%s", "library", image))
+	}
+
+	// The image doesn't seem to match a pattern we know
+	return nil, microerror.New(fmt.Sprintf("Unable to determine a registry path for image %s", image))
```
MarcelMue: We want to always use typed errors. Please do not use the `New` option like this. 

Read about it here: https://github.com/giantswarm/fmt/blob/master/go/errors.md#custom-errors

pkg/registry/registry_quay.go
```diff
@@ -0,0 +1,126 @@
...
+	"regexp"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/retagger/pkg/images"
+)
+
+// This code is largely lifted from nokia/docker-registry-client. It provides Quay-specific API behavior needed by retagger.
+
+var (
+	ErrNoMorePages = errors.New("No more pages")
```
MarcelMue: Same as above. This should be in an `errors.go` like elsewhere in our code. 

Same link: https://github.com/giantswarm/fmt/blob/master/go/errors.md#custom-errors

pkg/registry/registry_quay.go
```diff
@@ -0,0 +1,126 @@
...
+	"fmt"
+	"net/http"
+	nurl "net/url"
+	"regexp"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/retagger/pkg/images"
+)
+
+// This code is largely lifted from nokia/docker-registry-client. It provides Quay-specific API behavior needed by retagger.
```
MarcelMue: I understand this, but we should refactor it to fit better.

pkg/registry/registry_quay.go
```diff
@@ -0,0 +1,126 @@
...
+
+// Matches an RFC 5988 (https://tools.ietf.org/html/rfc5988#section-5)
+// Link header. For example,
+//
+//    <http://registry.example.com/v2/_catalog?n=5&last=tag5>; type="application/json"; rel="next"
+//
+// The URL is _supposed_ to be wrapped by angle brackets `< ... >`,
+// but e.g., quay.io does not include them. Similarly, params like
+// `rel="next"` may not have quoted values in the wild.
+var nextLinkRE = regexp.MustCompile(`^ *<?([^;>]+)>? *(?:;[^;]*)*; *rel="?next"?(?:;.*)?`)
```
MarcelMue: This should be up top.

pkg/registry/registry.go
stone-z: I agree, but I think it holds and I can't think of a more automatic way to resolve these names. We don't do this elsewhere because before we let docker handle this resolution when we pulled the image. This is for handling the three ways an image might be listed in the config:
- `some.domain.com/some/full/path` --> FQDN case
- `some/image` --> Docker hub case --> exactly two parts
- `someimage` --> Docker library case --> exactly one part

I don't think it's possible for `some/longer/path` to be a valid docker image unless it also specifies the host

pkg/registry/registry_quay.go
stone-z: Agree entirely

pkg/registry/registry_quay.go
stone-z: Done

pkg/registry/registry_quay.go
stone-z: Wrapped

pkg/registry/registry.go
stone-z: Wrapped

pkg/registry/registry.go
stone-z: Good point

pkg/images/images.go
stone-z: Done

pkg/images/images.go
stone-z: Done

cmd/flag.go
stone-z: Changed

pkg/registry/registry.go
stone-z: I found a much better way to do this today, will replace this

pkg/registry/error.go
```diff
@@ -68,3 +69,12 @@ func IsRepositoryNotFound(err error) bool {
 		return false
 	}
 }
+
+var noMorePagesError = &microerror.Error{
+	Kind: "noMorePagesError",
+}
+
+// IsNoMorePages asserts noMorePagesError.
+func IsNoMorePages(err error) bool {
```
MarcelMue: This error should come before `IsRepositoryNotFound` - `RepositoryNotFound` is after `NoMorePages` in the alphabet #nitpick

pkg/registry/registry_quay.go
```diff
@@ -0,0 +1,125 @@
...
+
+// GetQuayTagsWithDetails fetches tags for the given image including extra information defined in a QuayTag
+// This uses the Quay API, so assumes a Quay host. Other hosts are likely to fail.
+func (r *Registry) GetQuayTagsWithDetails(image string) (tags []QuayTag, err error) {
+	url := fmt.Sprintf("https://%s/api/v1/repository/%s/tag/", r.host, images.Name(r.organisation, image))
+
+	var response TagsResponse
+	for {
+
```
MarcelMue:
```suggestion
```

pkg/registry/registry_quay.go
```diff
@@ -0,0 +1,125 @@
...
+		r.logger.Log("level", "debug", "message", fmt.Sprintf("requesting registry tags from %s", url))
+
+		url, err = r.getPaginatedJSON(url, &response)
+		if err != nil {
+			if IsNoMorePages(err) {
+				tags = append(tags, response.Tags...)
+				return tags, nil
+			}
+
+			return nil, err
```
MarcelMue: We usually put this in an `else` statement after matching the explicit error - not a big deal though.

pkg/retagger/executor.go
```diff
@@ -21,7 +29,148 @@ func (r *Retagger) ExecuteJobs() error {
 	return nil
 }
 
+// executeJob accepts a job definition and runs it as either a pattern job or a single job
```
MarcelMue:
```suggestion
// executeJob accepts a job definition and runs it as either a pattern job or a single job.
```

pkg/retagger/executor.go
```diff
@@ -21,7 +29,148 @@ func (r *Retagger) ExecuteJobs() error {
...
 func (r *Retagger) executeJob(job Job) error {
+	if job.SourceSha != "" {
+		if job.SourcePattern != "" {
+			// Configuration specified a SHA and a pattern -- use SHA to be safe, but warn about misconfiguration
```
MarcelMue:
```suggestion
			// Configuration specified a SHA and a pattern -- use SHA to be safe, but warn about misconfiguration.
```

pkg/retagger/executor.go
```diff
@@ -21,7 +29,148 @@ func (r *Retagger) ExecuteJobs() error {
...
+// executePatternJob accepts a job definition containing a tag pattern
+// and creates and runs single jobs for each tag matching the pattern.
+func (r *Retagger) executePatternJob(job Job) error {
+	jobs, err := r.compilePatternJobs(job)
+	if err != nil {
+		return err
+	}
+
+	for _, job := range jobs {
+		r.executeSingleJob(job, false) // Do not skip existing tags for this job
```
MarcelMue:
```suggestion
		// Do not skip existing tags for this job.
		r.executeSingleJob(job, false)
```

pkg/retagger/executor.go
```diff
@@ -21,7 +29,148 @@ func (r *Retagger) ExecuteJobs() error {
...
+
+	return nil
+}
+
+// compilePatternJobs gets tag lists from the upstream repo, finds tags matching the pattern specified in the job,
+// and creates new jobs to retag new images and images which have changed since their last collection.
+func (r *Retagger) compilePatternJobs(job Job) ([]Job, error) {
+	r.logger.Log("level", "debug", "message", fmt.Sprintf("compiling jobs for image %v using pattern %v, with options %#v", job.SourceImage, job.SourcePattern, job.Options))
+
+	// Make sure our pattern is valid
```
MarcelMue:
```suggestion
	// Make sure our pattern is valid.
```

You get the _point_ for the rest of these :P

pkg/retagger/executor.go
```diff
@@ -21,7 +29,148 @@ func (r *Retagger) ExecuteJobs() error {
...
+				// Tag exists, but we should update the image
+
+				newDigest, err := externalRegistry.ManifestDigest(fullImageName, tag.Name)
+				if err != nil {
+					return nil, microerror.Mask(err)
+				}
+
+				if newDigest.String() != tag.ManifestDigest {
+					// Retag this image with this tag
+					r.logger.Log("level", "debug", "message",
```
MarcelMue: I am not a big fan of splitting lines here. It would make more sense to me to have the `fmt` foo in its own line with a variable instead. 

Personal preference though.

pkg/retagger/job.go
```diff
@@ -20,6 +21,9 @@ type JobOptions struct {
 	TagSuffix string
 
 	OverrideRepoName string
+
+	// UpdateOnChange sets whether a pattern Job should update the destination image if a source image changes for a given tag
```
MarcelMue: I think we would _nearly_ never want to do this. We have retagger because we want our images to be stable and to guarantee that they don't move. The option to move a tag should therefor never be included.

WDYT?

stone-z: The main use I anticipate would be for tags like `latest` or `stable`, for example.

pkg/retagger/job_pattern.go
```diff
@@ -66,40 +66,43 @@ func (job *PatternJob) Compile(r *Retagger) ([]SingleJob, error) {
 	jobs := []SingleJob{}
 
 	for _, match := range matches {
-		sourceSHA := ""
+		// sourceSHA := ""
```
MarcelMue: Can be removed?

pkg/retagger/job.go
MarcelMue: Are there cases where we would want to retag `latest` or `stable`? I think we should always work against immutable tags here. We generally don't want to follow moving tags to keep everything reproducible / testable and so on.

pkg/registry/registry.go
```diff
@@ -2,15 +2,17 @@ package registry
...
 
+	dockerRef "github.com/docker/distribution/reference"
 	dockerclient "github.com/docker/docker/client"
-	"github.com/giantswarm/backoff"
-	"github.com/giantswarm/microerror"
-	"github.com/giantswarm/micrologger"
 	"github.com/nokia/docker-registry-client/registry"
 	"github.com/opencontainers/go-digest"
 
+	"github.com/giantswarm/backoff"
```
MarcelMue: Not sure where this is coming from: We generally have 3 blocks: 
1. Standard lib imports
2. External imports
3. Project internal imports

So these imports belong into the second block:
https://github.com/giantswarm/fmt/blob/master/go/imports.md

pkg/retagger/job.go
stone-z: I don't like moving tags, but I see two reasons we might want it. The first is for automatically testing images which publish `dev` or `latest` or whatever but not tags for pre-release versions (so not hyperkube, but smaller stuff). Like "can we run this latest app version?" We can also just agree to never do this.

The bigger thing is that we _should_ be informed if a tag does actually change upstream because if an otherwise "fixed" tag gets moved it is probably either a big mistake or a very serious reason.

I agree we always want to use fixed version numbers everywhere else

pkg/retagger/job_pattern.go
stone-z: Whoops. Yes

pkg/registry/registry.go
stone-z: VS import organization /shrug

pkg/retagger/executor.go
stone-z: I un-split it. Looks nicer

pkg/retagger/executor.go
stone-z: _Point_ taken

pkg/retagger/job.go
stone-z: So as it turns out, the image ID reported by Quay is not the actual image ID, making the check for changed digests actually useless. If we want to check for changed tags, we will either need to pull every image again (obviously not worth it) or use some secondary storage where we keep the SHA we pushed to Quay at acquisition and compare it on subsequent checks. That actually isn't so bad, but I'm not sure if it's worth it. So in light of that, I think it is ok to remove the Update feature

pkg/retagger/job_single.go
```diff
@@ -0,0 +1,108 @@
...
+func getFullImageNameForJob(j *JobDefinition, r *Retagger) string {
+	registryPath, err := r.registry.GuessRegistryPath(j.SourceImage)
+	if err != nil {
+		return j.SourceImage // Fall back to trying to use given image name.
+	}
+	return strings.Trim(registryPath.Path, "/") // Remove leading slash.
+}
+
+// GetSourceForJob populates a Source object based on the given JobDefinition.
+func GetSourceForJob(jobDef *JobDefinition, r *Retagger) Source {
```
MarcelMue: The order in this file is all over the place. It's common in go to have public methods before private ones. We also usually follow the convention of having methods before functions. I would suggest the same here.

pkg/retagger/executor.go
```diff
@@ -6,76 +6,86 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
-func (r *Retagger) ExecuteJobs() error {
-	r.logger.Log("level", "debug", "message", fmt.Sprintf("start executing %d jobs", len(r.jobs)))
+// CompileJobs compiles all jobs in this Retagger's list of jobs into a list of concrete SingleJobs.
+func (r *Retagger) CompileJobs() error {
+	r.logger.Log("level", "debug", "message", fmt.Sprintf("compiling %d job definitions", len(r.jobs)))
 
+	// TODO: Gate for not performing API calls during compile?
```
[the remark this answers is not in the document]
stone-z: Removed. I considered whether there would be a case we'd want to do a dry-run without actually fetching tag lists from external registries (so using no network or only network to our quay repos). I don't think we need it

pkg/retagger/job_single.go
stone-z: Sorted

pkg/retagger/retagger.go
```diff
@@ -8,18 +8,26 @@ import (
...
 type Retagger struct {
 	logger   micrologger.Logger
 	registry *registry.Registry
+	dryrun   bool
 
-	jobs []Job
+	jobs         []JobDefinition
+	compiledJobs []SingleJob
+
+	CurrentTags map[string]map[string]registry.QuayTag
```
MarcelMue:
```suggestion
	currentTags map[string]map[string]registry.QuayTag
```

README.md
```diff
@@ -29,8 +29,13 @@ will become
...
-expect.
+Images can be defined using their SHA or using a tag pattern.
+
+If using a SHA, the exact image version is pulled from the source repository, pushed to Quay, and tagged with the given tag.
+
+If using a pattern, all tags in the original repository are compared against the pattern.
+If a matching tag is found, the image at that tag is pulled and retagged in Quay with the same tag
+(e.g. `alpine:9.99` matches pattern `9.[0-9]*` --> retagged as `quay.io/giantswarm/alpine:9.99`).
 
 Besides avoiding manual work, we do this for accountability and reproducability.
```
tfussell: Typo: reproducibility

integration/test/e2e/e2e_test.go
```diff
@@ -45,6 +45,7 @@ func TestE2e(t *testing.T) {
 		t.Fatal(err)
 	}
```
tfussell: Since 2.5 doesn't exist, maybe it would be worth adding a test case like `CheckImageDoesntExistOrFail`.

integration/test/e2e/e2e_test.go
```diff
@@ -45,6 +45,7 @@ func TestE2e(t *testing.T) {
...
 
+	CheckImageExistOrFail(t, r, "retagger-e2e", "2.6")
 	CheckImageExistOrFail(t, r, "retagger-e2e", "3.7")
```
tfussell: It might be time to turn these into a [table test](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

pkg/retagger/retagger.go
```diff
@@ -29,13 +37,31 @@ func New(config Config) (*Retagger, error) {
...
+		logger:      config.Logger,
+		registry:    config.Registry,
+		dryrun:      config.DryRun,
+		CurrentTags: map[string]map[string]registry.QuayTag{},
 	}
 	return r, nil
 }
 
+// GetTagDetails retrieves and caches the map of tag name: tag details for the given image.
+func (r *Retagger) GetTagDetails(image string) (map[string]registry.QuayTag, error) {
```
tfussell: I think this could be made package private.

pkg/retagger/job_single.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// Execute runs the job using the given Retagger instance.
+func (job *SingleJob) Execute(r *Retagger) error {
+	return r.executeSingleJob(*job)
+}
+
+// ShouldRetag examines the state of the job and the given Retagger's registry and returns whether the job should be run.
+func (job *SingleJob) ShouldRetag(r *Retagger) (bool, error) {
+
```
tfussell: No need for a blank first line.

pkg/retagger/job_single.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/giantswarm/microerror"
+)
+
+// SingleJob is a concrete job which can be executed.
+type SingleJob struct {
+	Source Source
+
+	Destination Destination
+
```
tfussell: If there's meaning to this whitespace, I think it should have a comment and if not, let's get rid of it.

pkg/retagger/job_pattern.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	// Get the tags for this image from the external registry.
+	externalRegistryTags, err := r.Tags(image)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	// Find tags matching our configured pattern.
+	matches := []string{}
```
tfussell: I would prefer `var matches []string` here. Go allows you to append to a `nil` slice and it avoid unnecessary initialization.

pkg/retagger/job_pattern.go
```diff
@@ -0,0 +1,130 @@
...
+	// Get SHA/Tag pairs from our quay registry.
+	quayTagMap, err := r.GetTagDetails(job.Destination.Image)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	jobs := []SingleJob{}
+
+	for _, match := range matches {
+
```
tfussell: Extraneous whitespace here.

tfussell:
Left a few small comments. Excited to get this running! Might be time to open it up to more teams for review once these are resolved.

pkg/retagger/job_pattern.go
stone-z: I like that much better. Changed in a couple other places as well

integration/test/e2e/e2e_test.go
stone-z: Added

pkg/retagger/job_definition.go
```diff
@@ -0,0 +1,167 @@
...
+func (jr *JobDefinition) Compile(r *Retagger) ([]SingleJob, error) {
+	job, err := jr.toSingleOrPatternJob(r)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	return job.Compile(r)
+}
+
+func (jr *JobDefinition) toSingleOrPatternJob(r *Retagger) (CompilableJob, error) {
```
MarcelMue: Public functions then private ones.

PR #299, opened 2019-12-12
add loki images

piontec:
Add loki v1.2.0 images


rossf7:
@piontec Thanks for fixing the order. We are pretty militant about ordering but I didn't know we had it in CI. 😂 

No need for an approval on retagger PRs. A ping is fine.

https://github.com/giantswarm/giantswarm/blob/master/processes/merging-without-approval.md#retagger-prs

piontec:
> If you are just adding a new tag to an existing image you do not need to wait for approval.

It's a new image :P But sure :)

rossf7:
> It's a new image 

Good catch and I wrote those docs! :)

PR #303, opened 2019-12-29
Add kindest/node 1.16.3

tfussell:
For e2ectl to set up 1.16.3 clusters.


tfussell:
It seems that e2ectl uses the latest tag as the default cluster version (https://github.com/giantswarm/e2ectl/blob/master/cmd/cluster/create/flag.go#L37). Is it okay to go ahead and merge or should we wait so e2e tests don't break?

PR #310, opened 2020-01-06
Retag Heptio EventRouter

omegas27:
_Retagger_

Needed for https://github.com/giantswarm/eventrouter-app/tree/add-eventrouter/helm/eventrouter-app as the image does not define a user

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -93,6 +93,15 @@
   tags:
   - sha: 59eec8837a4d942cc19a52b8c09ea75121acc38114a2c68b98983ce9356b8610
     tag: "3.1"
+- name: gcr.io/heptio-images/eventrouter
+  tags:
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -338,6 +338,15 @@
   tags:
   - sha: 97697df680ea4c7e70c4cb4af5bdc44f7a70b25be7afde70bd921a658e4c62ec
     tag: v0.30.1
+- name: gcr.io/heptio-images/eventrouter
```
MarcelMue: argh sorting needs to be alphabetical

omegas27: Do I sort by repo or image name?

MarcelMue: Repo then image - so it should be up there with the other `gcr.io` ones

omegas27: Thanks :) It should be better now :)

PR #311, opened 2020-01-06
Add root user to eventrouter retag

omegas27:
Added root user to be able to run retag https://circleci.com/gh/giantswarm/retagger/1545


omegas27:
It switched to the root user in the dockertile to create the giantswarm user and then switch to the newly created giantswarm user.
Since it is to be used in a managed app, we prefer to not let it run as root

PR #317, opened 2020-01-09
Fix aliyun tags list

stone-z:
Provides the basic functionality needed to support pattern jobs on Aliyun. This will need to be refactored, but will un-break `retagger` for Aliyun in the meantime


pkg/registry/registry_quay.go
```diff
@@ -50,6 +66,11 @@ var nextLinkRE = regexp.MustCompile(`^ *<?([^;>]+)>? *(?:;[^;]*)*; *rel="?next"?
 // GetQuayTagsWithDetails fetches tags for the given image including extra information defined in a QuayTag
 // This uses the Quay API, so assumes a Quay host. Other hosts are likely to fail.
 func (r *Registry) GetQuayTagsWithDetails(image string) (tags []QuayTag, err error) {
+	if r.host == "registry-intl.cn-shanghai.aliyuncs.com" {
```
MarcelMue: Something in the logic seems wonky if this is needed - why not create the correct registry struct right away?

pkg/registry/registry_aliyun.go
```diff
@@ -0,0 +1,118 @@
...
+func (t *AliyunTag) GetDigest() string {
+	return t.Digest
+}
+
+func (t *AliyunTag) GetSize() int64 {
+	return t.ImageSize
+}
+
+func (r *Registry) GetAliyunTagsWithDetails(image string) (tags []QuayTag, err error) {
+	crClient, err := cr.NewClientWithAccessKey("cn-shanghai", r.accessKey, r.accessSecret)
```
MarcelMue: `cn-shanghai` should be in a `key.go` as it's a pretty important magic string here :D

cmd/flag.go
```diff
@@ -20,6 +22,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVarP(&f.Organization, "organization", "o", "giantswarm", "organization to tag images for")
 	cmd.Flags().StringVarP(&f.Username, "username", "u", "", "username to authenticate against registry")
 	cmd.Flags().StringVarP(&f.Password, "password", "p", "", "password to authenticate against registry")
+	cmd.Flags().StringVarP(&f.AccessKey, "access-key", "", "", "access key for registry api")
```
MarcelMue: Put these beauties up top to keep the alphabetical order.

pkg/registry/registry_quay.go
stone-z: Yep, this is quick and dirty just to get unblocked. Moving to separate backend-aware registry behavior is the likely goal but it's a bigger change and will take me a bit longer

pkg/registry/registry_aliyun.go
stone-z: This will be parameterized instead

pkg/registry/registry_aliyun.go
```diff
@@ -64,7 +65,7 @@ func (r *Registry) GetAliyunTagsWithDetails(image string) (tags []QuayTag, err e
 	}
 
 	tagRequest := cr.CreateGetRepoTagsRequest()
-	tagRequest.Domain = "cr.cn-shanghai.aliyuncs.com"
+	tagRequest.Domain = fmt.Sprintf("cr.%s.aliyuncs.com", r.aliyunRegion)
```
MarcelMue: Let's turn this `fmt.Sprintf` into a little function in `key.go` in a follow up PR.

cmd/flag.go
stone-z: Done

stone-z:
Issue for cleanup and redesign: https://github.com/giantswarm/giantswarm/issues/8306

PR #325, opened 2020-01-16
Go modules

stone-z:
Move to Go modules. First time I've done this so let me know if I missed anything


tfussell:
3.6 million deleted lines, not bad.

tfussell:
If I understand correctly, this should maybe be switched from `architect` to `architect-orb` which is able to, for example, ensure that `go mod tidy` doesn't change the `go.mod` file. More info here https://github.com/giantswarm/architect-orb/blob/master/README.md

stone-z:
I'd also be inclined to use `architect-orb`, but either the build job has to be able to persist the dependencies it fetched or we'll have to re-fetch them in the `compile e2e test` step. This currently fails because they're missing. Maybe there's a clean CircleCI way to do this? Ideally we wouldn't have to pull them again. I've for some feedback in #sig-releng

stone-z:
I would like to go ahead with this PR and revisit moving to `orb` in a separate PR

kopiczko:
`rettager` binary will be persisted to workspace by `architect/go-build` job https://circleci.com/orbs/registry/orb/giantswarm/architect#jobs-go-build.

stone-z:
> `rettager` binary will be persisted to workspace by `architect/go-build` job https://circleci.com/orbs/registry/orb/giantswarm/architect#jobs-go-build.

We also need the dependencies for it which are collected during build. Otherwise `go test -c` can't compile the e2e tests. Can we somehow tell the `go-build` job to keep these as well?

kopiczko:
@stone-z I never tried to use command (as opposed to job) but I think it's possible. You could try to use `architect/go-cache-restore` command for that which architect orb uses internally https://circleci.com/orbs/registry/orb/giantswarm/architect#commands-go-cache-restore.

PR #330, opened 2020-01-21
add kube-state-metrics v1.9.2

sslavic:
As preparation for upgrade of our app this PR add retagging for kube-state-metrics v1.9.2


sslavic:
Going with a ping

PR #331, opened 2020-01-21
add addon-resizer 1.8.7

sslavic:
Upgrading kube-state-metrics to 1.9.2 and noticed that there's newer addon-resizer too (used as sidecar in kube-state-metrics-app). addon-resizer 1.8.7 switches from extensions to to apps/v1 Deployments API making it k8s 1.16 compatible https://github.com/kubernetes/autoscaler/commits/addon-resizer-release-1.8/addon-resizer

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -205,8 +205,8 @@
     tag: 0.1
 - name: k8s.gcr.io/addon-resizer
   tags:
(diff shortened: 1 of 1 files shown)


sslavic:
Going with a ping here too

PR #334, opened 2020-01-22
Add cluster-autoscaler v1.17.0

sslavic:


images.yaml
```diff
@@ -209,20 +209,8 @@
...
-  - sha: eea97bdf90e3f6074af2fdeb3df40969c44a721dc8f4d148844de85fe93508ff
-    tag: v1.15.1
-  - sha: 5a80893ea1a16a066889ab06934cfa3996d87a7232b10554de4ed8788b1fa212
-    tag: v1.15.2
-  - sha: 8d8664ac086f5dcff5b95874ef8b1dfaab6eb87477bbaea9a2e19dfac063c521
-    tag: v1.16.2
-  - sha: 6b887823b1fd2789aa926f7bd3d5efa5c3bec5a194f72da121c292d95204a3b7
-    tag: v1.16.3
+  - sha: 69413a60c3764eb153098fb53ae1e5f0a9652cd6ccb1b3450d37cb9390d89ab3
+    tag: v1.17.0
```
MarcelMue: This is only recommended for k8s 1.17 - we are pretty far away from that. It therefor shouldn't be used in any current releases yet.

https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#releases

sslavic: Yes, still retagging 1.17.0 won't harm, won't be used anywhere yet.

v1.16.3 is already retagged so will bump our cluster-autoscaler-app to it and apply more RBAC changes in PR there.

PR #337, opened 2020-01-29
Add ingress-nginx 0.28.0

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/8664

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -404,8 +404,8 @@
     tag: v0.2.5
 - name: quay.io/kubernetes-ingress-controller/nginx-ingress-controller
   tags:
-  - sha: 3d93f5ba697f7014e0b639eb5b2abbaa7e20d70a4fd759612c52c1fc5d506612
-    tag: 0.27.1
+  - sha: ca2eee26afd16dc052c7950f13df6c906be279de64d990e88383c9f123556e06
+    tag: 0.28.0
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -404,8 +404,8 @@
     tag: v0.2.5
 - name: quay.io/kubernetes-ingress-controller/nginx-ingress-controller
   tags:
-  - sha: 3d93f5ba697f7014e0b639eb5b2abbaa7e20d70a4fd759612c52c1fc5d506612
-    tag: 0.27.1
+  - sha: ca2eee26afd16dc052c7950f13df6c906be279de64d990e88383c9f123556e06
+    tag: 0.28.0
```
MarcelMue: Would be neat to add a pattern like we did for hyperkube. That way you don't need to PR for this anymore. @giantswarm/team-ludacris Should help / is happy to help.

sslavic: Thanks, didn't notice before. Is pulling smart, only to pull the missing images?

MarcelMue: It will only check before pulling AFAIK

Here it would be good to have a regex that matches everything greater than 0.28.0 though - but that regex becomes complex quickly :D

sslavic: When I figure it out will add the pattern, thanks once more for heads up

PR #340, opened 2020-01-30
Replace images with patterns

MarcelMue:


README.md
```diff
@@ -111,19 +111,15 @@ the target tag should be `v1.5.2-2`, `v1.5.2-3`, and so on.
...
-
 For example, at the time of this writing:
 
 ```yaml
 - name: k8s.gcr.io/hyperkube
   patterns:
-    - pattern: '\A(v1.17.[0-9]+)(-rc.[0-9.]+)*\z'       # Match any v1.17.x or v1.17.x-rc.x              -- matches 3 tags
-    - pattern: '\A(v[0-9.]+)\z'                         # Match any vx.x.x tag, excluding alphas, etc.   -- matches 193 tags
-    - pattern: '\A(v[0-9.]+)(-(alpha|beta)[0-9.]+)*\z'  # Match any version, including alphas and betas  -- matches 485 tags
+    - pattern: 'v1.17.x'       # Match any v1.17.x
```
tfussell: `v1.17.x` syntax is equivalent to `>= v1.17.0`, and this might be clearer as an example that you don't need to specify exact versions.

README.md
```diff
@@ -111,19 +111,15 @@ the target tag should be `v1.5.2-2`, `v1.5.2-3`, and so on.
 
 It is also possible to watch and automatically retag new tagged releases in the upstream repository.
 To do this, specify a pattern for the image in the `images.yaml` configuration file.
-Each pattern must be a valid Go regular expression (you can try it out [here](https://regex-golang.appspot.com/assets/html/index.html)),
+Each pattern must be a valid semver expression (you can read about it [here](https://github.com/Masterminds/semver)),
```
tfussell: Maybe mention that the `v` is optional. Or for consistency we could require that it is there.

README.md
```diff
@@ -82,7 +82,7 @@ What the attributes mean:
 - `patterns[].pattern`: Valid Go regular expression to match tags.
```
tfussell: This should be updated too.

MarcelMue: All done - merging here. Ping me if anything is still wrong.

PR #342, opened 2020-01-31
Try registry backoff

tfussell:
To avoid registry rate limiting, this creates a Registry manually instead of using registry.New and passes in an http.Client with a custom transport wrapper which applies exponential backoff in the case of a 429 response.

### DIFF
--- a/pkg/retagger/job_pattern.go
+++ b/pkg/retagger/job_pattern.go
@@ -84,7 +84,7 @@ func (job *PatternJob) Compile(r *Retagger) ([]SingleJob, error) {
 				logger: r.logger,
 			},
 		},
-		Logf: dockerRegistry.Quiet,
+		Logf: dockerRegistry.Log,
 		URL:  fmt.Sprintf("https://%s", job.Source.RepoPath),
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/retagger/job_pattern.go
```diff
@@ -22,18 +27,65 @@ type PatternJob struct {
...
+			var respErr error
+			resp, respErr = t.Transport.RoundTrip(request)
+			// Internal error, return nil to prevent retry
+			if respErr != nil {
+				err = respErr
+				return nil
+			}
+			// Rate limited
+			if resp.StatusCode == 429 {
+				return microerror.New("rate limited")
```
stone-z: Need an error type to match here?

stone-z:
I can't tell from looking at the code but will this affect pagination behavior? Looks like CI is failing because it can't find 2.6 but it does still exist on docker hub. Maybe pages are missing

tfussell:
> I can't tell from looking at the code but will this affect pagination behavior? Looks like CI is failing because it can't find 2.6 but it does still exist on docker hub. Maybe pages are missing

I hadn't set up the transport wrapping correctly. This appears to be fixed now.

pkg/retagger/job_pattern.go
tfussell: Good call. This was a placeholder. I might have forgotten about it. Fixing now.

tfussell:
Merging after e2e passes.

PR #349, opened 2020-02-06
Check aliyun repo exists

stone-z:
Aliyun doesn't actually store the error code in the ErrorCode field, so search for it in the message.

Towards https://github.com/giantswarm/giantswarm/issues/8807

### DIFF
--- a/pkg/registry/registry_aliyun.go
+++ b/pkg/registry/registry_aliyun.go
@@ -83,7 +83,8 @@ func (r *Registry) GetAliyunTagsWithDetails(image string) (tags []QuayTag, err e
 				r.logger.Log("level", "warn", "message", "Aliyun repository does not exist. Retagger will try to create it")
 				return aliTags, nil
 			}
-		} else if err != nil {
+		}
+		if err != nil {
 			return nil, microerror.Mask(err)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/registry/registry_aliyun.go
```diff
@@ -76,6 +77,13 @@ func (r *Registry) GetAliyunTagsWithDetails(image string) (tags []QuayTag, err e
 		tagRequest.Page = requests.NewInteger(page)
 
 		crResp, err := crClient.GetRepoTags(tagRequest)
+		serverError, isAliyunError := err.(*errors.ServerError)
+		if isAliyunError {
+			if strings.Contains(serverError.Message(), "REPO_NOT_EXIST") {
+				r.logger.Log("level", "warn", "message", "Aliyun repository does not exist. Retagger will try to create it")
```
tfussell: It seems like this just make the error message a little nicer. Missing repo creation hasn't been implemented, has it?

MarcelMue: This was done here:
https://github.com/giantswarm/retagger/pull/287

So we only need to prevent actually erroring out here.

tfussell: Totally forgot about that. Thanks.

PR #355, opened 2020-02-08
Retry on trailer EOF error

tfussell:
Solves https://github.com/giantswarm/giantswarm/issues/8841

If a chunked response ends unexpectedly (maybe due to a server failure, network issues, or silent rate limiting), the Go http transport will attempt to parse trailers (like headers but at the end of a response), find zero bytes remaining instead of `\r\n` (indicates the final chunk) and return a new error with string `"http: unexpected EOF reading trailer"`. This error happens intermittently, maybe 1% of the time during a full retagger execution. My solution is to wrap the registry tags function call with a backoff retry just like for the rate limiting issue. I also cleaned up the rate limiting backoff using `backoff.Permanent` which I just discovered.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/docker/go-units v0.4.0 // indirect
 	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7 // indirect
 	github.com/giantswarm/backoff v0.0.0-20190913091243-4dd491125192
-	github.com/giantswarm/microerror v0.0.0-20191011121515-e0ebc4ecf5a5
+	github.com/giantswarm/microerror v0.1.0
 	github.com/giantswarm/micrologger v0.0.0-20191014091141-d866337f7393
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/docker/go-units v0.4.0 // indirect
 	github.com/docker/libtrust v0.0.0-20160708172513-aabc10ec26b7 // indirect
 	github.com/giantswarm/backoff v0.0.0-20190913091243-4dd491125192
-	github.com/giantswarm/microerror v0.0.0-20191011121515-e0ebc4ecf5a5
+	github.com/giantswarm/microerror v0.1.0
```
MarcelMue: Is this the correct version for modules now?

tfussell: `v0.1.0` is the last release using dep. Since the go-modules and errors package PRs haven't been merged yet, this seemed like a good target.

PR #371, opened 2020-02-20
Add image for golangci-lint

stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/8901 https://github.com/giantswarm/giantswarm/issues/7658

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -175,6 +175,10 @@
 - name: golang
   patterns:
   - pattern: '>= 1.13.1'
+- name: golangci/golangci-lint
+  tags:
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -175,6 +175,10 @@
 - name: golang
   patterns:
   - pattern: '>= 1.13.1'
+- name: golangci/golangci-lint
+  tags:
+  - sha: 757ee77f748bb5d93d584a9ab5710f4149adcdacf2f3584c14b42015f7811b64
+    tag: v1.23.6
```
MarcelMue: Hold on - why not use a pattern? Pretty sure this evolves with the go version, right?

stone-z: Following my own advice and not using a pattern until there is some automation to actually use a new image -- stay tuned

PR #384, opened 2020-03-03
Tag golang alpine prereleases

tfussell:
Towards https://github.com/giantswarm/architect/pull/367

Golang alpine tags are considered as prereleases so they're not handled well by semver so I'm going with the old fashioned tag approach.


images.yaml
```diff
@@ -178,7 +178,7 @@
     tag: v2.6.2-v1.6.0
 - name: golang
   patterns:
-  - pattern: '>= 1.13.1'
+  - pattern: '>= 1.13.1-alpine3.10'
```
stone-z: Given that semver didn't realize this wasn't a prerelease, is it possible this would include future/other prereleases? Running locally with --dry-run I get 
`D 03/04 03:52:48 Found 124 upstream tags which match the pattern >= 1.13.1-alpine3.10 | retagger/pkg/retagger/job_pattern.go:52`

PR #404, opened 2020-03-23
Adjust Kibana 7.6.1

T-Kukawka:
_Retagger_

---


T-Kukawka:
@webwurst it was used on old efk - so i assumed it is obsolete

PR #359, opened 2020-02-11
Add image for Helm chart testing

yasn77:
Needed for https://github.com/giantswarm/architect-orb/pull/65

_Retagger_

https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/container-registry.md

---

PR #410, opened 2020-03-30
Add conftest docker image

fiunchinho:
Tagging `conftest` docker image so we can use it.

Quay repository here https://quay.io/repository/giantswarm/conftest


yasn77:
> Unsure if you have to setup the aliyun registry (for china) manually also, but better check this.
> https://github.com/giantswarm/giantswarm/blob/master/ops-recipes/container-registry.md#alibaba-registry

In #359 (https://github.com/giantswarm/retagger/pull/359#issuecomment-585260889) @MarcelMue  suggests it's not needed ¯\_(ツ)_/¯

PR #421, opened 2020-04-08
Add patterns for fluentd images

omegas27:
_Retagger_

---

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -75,8 +75,11 @@
   - sha: 0ae482ad3e8d951a66090b968389f2a11b0b16b8a208b5edb4ccb0ca5800b90d
     tag: "0.18"
 - name: fluent/fluentd-kubernetes-daemonset
-  patterns:
-  - pattern: '>= 1.x.x-debian-cloudwatch-1.0'
+  tags:
+  - sha: f9fa60dab5e7a6ae4adbe972a2fc22cf7eaa148df74cde4cd7f57e7019260451
+    tag: "v1.3-debian-cloudwatch"
+  - sha: da2c32a98e87809ecc6c2aea3be0d0b0f5cab31c4d75b6fb46b86d76d5554f23
+    tag: "v1.9.3-debian-cloudwatch-1.0"
 - name: gcr.io/google_containers/defaultbackend
   tags:
   - sha: ee3aa1187023d0197e3277833f19d9ef7df26cee805fef32663e06c7412239f9
--- a/images.yaml
+++ b/images.yaml
@@ -68,17 +68,15 @@
   - sha: 72d816c99cc86b117baf04385020121f71015d7f5f9eba2a59e72532d645d38f
     tag: 7.6.1
 - name: fluent/fluent-bit
-  tags:
-  - sha: 6861ee5ea81fbbacf8802c622d9305930b341acc0800bbf30205b4d74ce2b486
-    tag: 0.14.6
+  patterns:
+  - pattern: '>= 1.[0-9]+.[0-9]+'
 - name: fluent/fluent-bit-0.13-dev
   tags:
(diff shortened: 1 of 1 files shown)


axbarsan:
If I understand correctly, this will automatically update the fluentd images in our quay repo, when there's a new one released, right?

omegas27:
As long as they match the pattern, yes :) but it will also upload them to alyun in China

images.yaml
```diff
@@ -68,17 +68,15 @@
...
 - name: fluent/fluent-bit-0.13-dev
   tags:
   - sha: 0ae482ad3e8d951a66090b968389f2a11b0b16b8a208b5edb4ccb0ca5800b90d
     tag: "0.18"
 - name: fluent/fluentd-kubernetes-daemonset
-  tags:
-  - sha: f9fa60dab5e7a6ae4adbe972a2fc22cf7eaa148df74cde4cd7f57e7019260451
-    tag: "v1.3-debian-cloudwatch"
+  patterns:
+  - pattern: '>= 1.[0-9]+.[0-9]+-debian-cloudwatch-1.0'
```
MarcelMue, line 79: Careful with this one - we use a semver comparison and this might try to tag all pre release versions. Let me check.

omegas27: Maybe I can adjust the pattern to 1.[3-9][0-9]*.[0-9]+-debian-cloudwatch-1.0 ?

MarcelMue:
```
{"caller":"retagger/main.go:50","level":"error","message":"failed to execute command","stack":"[{/home/marcel/golang/src/github.com/giantswarm/retagger/cmd/runner.go:36: } {/home/marcel/golang/src/github.com/giantswarm/retagger/cmd/runner.go:98: } {/home/marcel/golang/src/github.com/giantswarm/retagger/pkg/retagger/executor.go:17: } {/home/marcel/golang/src/github.com/giantswarm/retagger/pkg/retagger/job_pattern.go:46: } {/home/marcel/golang/src/github.com/giantswarm/retagger/pkg/retagger/job_pattern.go:105: } {improper constraint: \u003e= 1.[0-9]+.[0-9]+}]","time":"2020-04-08T08:06:38.962915+00:00"}
```
I don't think semver likes this :(

MarcelMue:
See here:
https://github.com/Masterminds/semver#basic-comparisons

images.yaml
omegas27: Ok, so it is not a regex pattern. I guess 1.x.x would be better then

omegas27:
@stone-z I need it for an app. I will check batman or halo has some automation to update apps automatically

images.yaml
```diff
@@ -68,17 +68,15 @@
...
 - name: fluent/fluent-bit-0.13-dev
   tags:
   - sha: 0ae482ad3e8d951a66090b968389f2a11b0b16b8a208b5edb4ccb0ca5800b90d
     tag: "0.18"
 - name: fluent/fluentd-kubernetes-daemonset
-  tags:
-  - sha: f9fa60dab5e7a6ae4adbe972a2fc22cf7eaa148df74cde4cd7f57e7019260451
-    tag: "v1.3-debian-cloudwatch"
+  patterns:
+  - pattern: '>= 1.x.x-debian-cloudwatch-1.0'
```
MarcelMue, line 79: First one now works, second one it now tags all `pre-release` tags :
```
{"caller":"github.com/giantswarm/retagger/pkg/retagger/job_pattern.go:52","level":"debug","message":"Found 550 upstream tags which match the pattern \u003e= 1.x.x-debian-cloudwatch-1.0","time":"2020-04-08T08:39:49.395126+00:00"}
```
So maybe keep the first pattern and set them explicitly here until we come up with a solution for pre-releases.

omegas27: Thanks :)

PR #428, opened 2020-04-17
allow alpine non-root user to sudo

glitchcrab:
towards https://github.com/giantswarm/nginx-ingress-controller-app/pull/60/files

this allows the non-root user to use `sudo` to execute commands (provided this is allowed by the relevant pod security policy)

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -4,7 +4,11 @@
     customImages:
     - tagSuffix: giantswarm
       dockerfileOptions:
-      - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm && apk add --no-cache sudo && addgroup giantswarm wheel && echo "%wheel ALL=(ALL)  NOPASSWD: ALL" >> /etc/sudoers
+      - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -6,6 +6,10 @@
       dockerfileOptions:
       - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm
       - USER giantswarm
+    - tagSuffix: giantswarm-sysctl
+      dockerfileOptions:
+      - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm && apk add --no-cache sudo && echo "giantswarm  ALL = NOPASSWD: /sbin/sysctl" >> /etc/sudoers
```
corest, line 11: Can we split this sudo/echo part to a separate line RUN?

corest:
```suggestion
      - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm 
      - RUN apk add --no-cache sudo && echo "giantswarm  ALL = NOPASSWD: /sbin/sysctl" >> /etc/sudoers
```

glitchcrab: sure, i wasn't sure if retagger could handle that (didn't see any examples in the existing images)

PR #429, opened 2020-04-20
adds hugo and nginx

puja108:
needed for secure hugo deployment of internal docs

quay and aliyun repos created


puja108:
there was a strange version of oauth2 proxy already retagged, but I want the official upstream one, I guess as long as my tags don't overlap this should work, right?

puja108:
not sure why the image test is failing, what does that thing do?

MarcelMue:
`images are not sorted alphabetically` - a little hidden in the error message.

puja108:
oh, I have to sort it... now that is something that was not in docs

PR #434, opened 2020-05-01
Add component-specific images

tfussell:
Hyperkube is in the process of being deprecated so we should start retagging and using component-specific images such as kube-apiserver in our operators. We still need hyperkube for kubectl and kubelet until we find a better method so I haven't removed it.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -225,6 +225,18 @@
 - name: k8s.gcr.io/cluster-autoscaler
   patterns:
   - pattern: '>= v1.15.4'
+- name: k8s.gcr.io/kube-apiserver
+  patterns:
+  - pattern: '>= v1.16'
+- name: k8s.gcr.io/kube-controller-manager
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -228,6 +228,18 @@
 - name: k8s.gcr.io/hyperkube
   patterns:
   - pattern: '>= v1.15'
+- name: k8s.gcr.io/kube-apiserver
+  patterns:
+  - pattern: '>= v1.16'
```
MarcelMue, line 233: Should we make it `>= 1.17` or do you want to test a transition with these?

tfussell: I was thinking it would be cool to make this change in 1.16 so we're prepared for 1.17. I guess we really only need 1.16.9+ in that case though.

MarcelMue: I see - let's go with `1.16.9+` then

PR #436, opened 2020-05-12
Mask errors when execution fails due to reading or parsing yaml

stone-z:
Followup to https://github.com/giantswarm/retagger/pull/361

Adds context to errors when loading the config file

### DIFF
--- a/pkg/images/error.go
+++ b/pkg/images/error.go
@@ -5,8 +5,3 @@ import "github.com/giantswarm/microerror"
 var executionFailedError = &microerror.Error{
 	Kind: "executionFailedError",
 }
-
-// IsExecutionFailed asserts executionFailedError.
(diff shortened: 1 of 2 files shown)


pkg/images/error.go
```diff
@@ -0,0 +1,12 @@
...
+import "github.com/giantswarm/microerror"
+
+var executionFailedError = &microerror.Error{
+	Kind: "executionFailedError",
+}
+
+// IsExecutionFailed asserts executionFailedError.
+func IsExecutionFailed(err error) bool {
+	return microerror.Cause(err) == executionFailedError
+}
```
kopiczko, line 12: This is the only error w/o matcher. See https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors

```suggestion
```

PR #440, opened 2020-05-18
Add CA certs to alpine image

glitchcrab:
towards giantswarm/giantswarm/issues/6290

This PR adds ca-certs and bumps the alpine version to the latest version.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -1,6 +1,6 @@
 - name: alpine
   patterns:
-  - pattern: '>= 3.9'
+  - pattern: '>= 3.11'
     customImages:
     - tagSuffix: giantswarm
       dockerfileOptions:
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -1,9 +1,10 @@
 - name: alpine
   patterns:
-  - pattern: '>= 3.9'
+  - pattern: '>= 3.11'
     customImages:
     - tagSuffix: giantswarm
       dockerfileOptions:
+      - RUN apk add --no-cache ca-certificates
```
MarcelMue, line 7: @stone-z How does this work currently? Will this cause retagging of already existing tags?

stone-z: No - this won't be applied until a new tag matches the pattern

PR #445, opened 2020-05-20
Clean up of some unused images

omegas27:


images.yaml
```diff
@@ -96,29 +92,22 @@
...
 - name: gcr.io/google_containers/pause-amd64
   tags:
   - sha: 59eec8837a4d942cc19a52b8c09ea75121acc38114a2c68b98983ce9356b8610
     tag: "3.1"
 - name: gcr.io/heptio-images/eventrouter
-  tags:
-  - sha: dba60a88600d2d94fcd4c365e2931e54ae9aa495e94a924f80814e019b7ea046
-    tag: v0.3
+  patterns:
+  - pattern: '>= v0.3'
```
stone-z, line 101: Why a pattern here? Are we updating these automatically?

images.yaml
```diff
@@ -96,29 +92,22 @@
...
     customImages:
     - tagSuffix: giantswarm
       dockerfileOptions:
       - USER root
       - RUN addgroup -g 1000 -S giantswarm && adduser -u 1000 -S giantswarm -G giantswarm
       - USER giantswarm
 - name: gcr.io/heptio-images/gangway
+  patterns:
+  - pattern: '>= v3.1.0'
```
stone-z, line 110: Same here

images.yaml
omegas27: We are not updating them, so I can leave the tag. I was thinking it was better to have all versions since we not paying for storage and images anyway but maybe we can keep the tags for those. I just think we should have all versions so we do not wait on retagger PR as this could become a blocker.
@stone-z do you prefer me to switch to tags for those?

stone-z: Pattern jobs make every retagger run longer and potentially more prone to failing on aliyun. Yes we don't pay for storage currently, but we pay for CircleCI, and what's the point of collecting thousands of images we'll never deploy? As I said, not a blocker, just a preference of mine that we continue to use tags for slow-changing images without automation already in place

PR #449, opened 2020-06-04
Retag external-dns from eu.gcr.io/k8s-artifacts-prod instead of zalando teapot registry

sslavic:
Towards giantswarm/giantswarm#11356


sslavic:
Going with a ping

PR #450, opened 2020-06-08
Retag pause-amd64 from k8s.gcr.io facade

sslavic:
Beside switching to use facade instead of deprecated `gcr.io/google_containers`, PR also changes ratg configuration from retagging just a fixed 3.1 tag, to retag by pattern, >=3.1.

See https://groups.google.com/d/msg/kubernetes-dev/MkXnkTSJ_vs/7wvClbkxAgAJ


sslavic:
Going with a ping

PR #456, opened 2020-06-23
Add dockerfileOptions to include ca certs in kube-apiserver

MarcelMue:

---

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -232,17 +232,17 @@
   patterns:
   - pattern: '>= v1.16.8'
     customImages:
-      - tagSuffix: giantswarm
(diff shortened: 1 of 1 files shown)


MarcelMue:
I tested this locally and `certs` end up in `/etc/ssl/certs` so I assume that this works fine. Not sure how else I can test this.

whites11:
> I tested this locally and `certs` end up in `/etc/ssl/certs` so I assume that this works fine. Not sure how else I can test this.

Just build the image in one running cluster, change the static pod manifest and restart the kubelet (ensure oidc is configured)

PR #462, opened 2020-06-29
Retag new coredns releases by pattern

sslavic:


stone-z:
Hey @sslavic do you have some automation using these images? If not, we prefer to stick with specific image SHAs. Each pattern makes every run longer and CircleCI costs money

PR #471, opened 2020-07-17
Remove retagging k8s-migrator

sslavic:
See https://gigantic.slack.com/archives/C04TGHDEF/p1595015375299700?thread_ts=1594638116.236700&cid=C04TGHDEF


sslavic:
Going with a ping

rossf7:
@sslavic Thanks for removing.

We used k8s-migrator for migrating apps from manifests in k8scloudconfig to helm charts. AFAIK we're no longer using it. If we need it again we can migrate it to our app platform.

sslavic:
Thanks for heads up! Will archive then the k8s-migrator repo.

PR #476, opened 2020-07-30
Add busybox pattern

ced0ps:
https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/container-registry/

---

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -37,7 +37,6 @@
 - name: busybox
   tags:
   - sha: 9f1003c480699be56815db0f8146ad2e22efea85129b5b5983d0e0fb52d9ab70
-    tag: 1.31.0
   patterns:
   - pattern: '>= 1.31.0'
 - name: centos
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -38,6 +38,8 @@
   tags:
   - sha: 9f1003c480699be56815db0f8146ad2e22efea85129b5b5983d0e0fb52d9ab70
     tag: 1.31.0
+  patterns:
```
sslavic, line 41: IIUC 1.31.0 tag is included in the pattern, so should be safe to remove it.

sslavic: @ced0ps I meant whole tags section, not just tag key in tag definition

images.yaml
```diff
@@ -37,7 +37,8 @@
 - name: busybox
   tags:
   - sha: 9f1003c480699be56815db0f8146ad2e22efea85129b5b5983d0e0fb52d9ab70
```
ced0ps, line 39:
```suggestion
```

PR #484, opened 2020-09-23
update nginxinc/nginx-unprivileged:1.18-alpine sha & add hugo 0.75.1

ubergesundheit:
docs needs an upgrade. I want to use these container images.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -333,7 +333,7 @@
   tags:
   - sha: 2d55b56b04c06a4992729ad5430d2fae1a6833bd9c4a87a00173ff4563bbee40
     tag: 1.17-alpine
-  - sha: ae4131942b2e200dd06cbc52915b6972275ad56d1a37788661d65ead039d3cf3
+  - sha: 94da2c39b836871c1f15c49e150a5e7d9c8b24dcc5c8d6c557fee1300f2cbb0b
     tag: 1.18-alpine
 - name: omnition/opencensus-collector
   patterns:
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -333,7 +333,7 @@
   tags:
   - sha: 2d55b56b04c06a4992729ad5430d2fae1a6833bd9c4a87a00173ff4563bbee40
     tag: 1.17-alpine
-  - sha: ae4131942b2e200dd06cbc52915b6972275ad56d1a37788661d65ead039d3cf3
+  - sha: 94da2c39b836871c1f15c49e150a5e7d9c8b24dcc5c8d6c557fee1300f2cbb0b
```
stone-z, line 336: Why are we moving tags?

ubergesundheit: its newer. But if this is not how we do things, we can use the old sha.

stone-z: I see, it looks like upstream moves tags? That is unfortunate. Then yeah, I guess this would be the most consistent with them.

ubergesundheit: So whats your opinion? Leave the sha alone or use the newer one? Could you approve?

stone-z (approved):
Someone asked for semver tags [here](https://github.com/nginxinc/docker-nginx-unprivileged/issues/48#issuecomment-694952836). Would be great if that happened but until then this is ok with me

stone-z:
Someone asked for semver tags [here](https://github.com/nginxinc/docker-nginx-unprivileged/issues/48). Would be great if that happened but until then this is ok with me

PR #502, opened 2020-11-24
switch linkerd from patterns to specific tags

glitchcrab:
https://intranet.giantswarm.io/docs/support-and-ops/ops-recipes/container-registry/

---


glitchcrab:
gonna merge this in as CI is currently broken

PR #537, opened 2021-01-27
Replace hyperkube repo

tfussell:
I realized that we could generate the `hyperkube` image dynamically in `retagger` with only a few changes by watching an upstream kubernetes image to get the version and then use a custom dockerfile to build the image using that version by downloading binaries. This replaces the https://github.com/giantswarm/hyperkube repo and allows us to new versions as soon as they are released without the Dependabot and release rigamarole.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -374,6 +374,24 @@
     - tagSuffix: giantswarm
       dockerfileOptions:
       - RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+- name: k8s.gcr.io/kube-apiserver
+  tagTrimVersionPrefix: true
+  overrideRepoName: hyperkube
+  patterns:
+    - pattern: '>= v1.19.8 || >= v1.20.0-0'
+      customImages:
+        - dockerfileOptions:
+            - |
+              FROM quay.io/giantswarm/alpine:3.12.1 AS downloader
+              WORKDIR /tmp/hyperkube
+              COPY --from=0 /usr/local/bin/kube-apiserver /tmp/hyperkube/kube-apiserver
+              RUN KUBERNETES_VERSION=$(/tmp/hyperkube/kube-apiserver --version | grep Kubernetes | cut -d ' ' -f 2) && \
+              wget -t 5 https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubelet && \
+              wget -t 5 https://storage.googleapis.com/kubernetes-release/release/$KUBERNETES_VERSION/bin/linux/amd64/kubectl && \
+              chmod +x /tmp/hyperkube/kubelet /tmp/hyperkube/kubectl
+              FROM scratch
+              COPY --from=downloader /tmp/hyperkube/kubelet /kubelet
+              COPY --from=downloader /tmp/hyperkube/kubectl /kubectl
 - name: k8s.gcr.io/kube-controller-manager
   patterns:
   - pattern: '>= v1.16.8'
--- a/images.yaml
+++ b/images.yaml
@@ -378,20 +378,20 @@
   tagTrimVersionPrefix: true
   overrideRepoName: hyperkube
   patterns:
-    - pattern: '>= v1.19.8 || >= v1.20.0-0'
-      customImages:
-        - dockerfileOptions:
-            - |
-              FROM quay.io/giantswarm/alpine:3.12.1 AS downloader
-              WORKDIR /tmp/hyperkube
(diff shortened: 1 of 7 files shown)


images.yaml
```diff
@@ -374,6 +374,24 @@
...
       - RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+- name: k8s.gcr.io/kube-apiserver
+  tagTrimVersionPrefix: true
+  overrideRepoName: hyperkube
+  patterns:
+  - pattern: '>= v1.19.8 || >= v1.20.0-0'
+    customImages:
+    - dockerfileOptions:
+      - |
+        FROM quay.io/giantswarm/alpine:3.12.1 AS downloader
```
tfussell, line 385: Adapted from https://github.com/giantswarm/hyperkube/blob/master/Dockerfile

images.yaml
```diff
@@ -374,6 +374,24 @@
     - tagSuffix: giantswarm
       dockerfileOptions:
       - RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+- name: k8s.gcr.io/kube-apiserver
+  tagTrimVersionPrefix: true
+  overrideRepoName: hyperkube
+  patterns:
+  - pattern: '>= v1.19.8 || >= v1.20.0-0'
```
tfussell, line 381: Starting with 1.20 I think it would be cool to get pre-release versions so we can get ahead of the release curve.

images.yaml
```diff
@@ -374,6 +374,24 @@
     - tagSuffix: giantswarm
       dockerfileOptions:
       - RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+- name: k8s.gcr.io/kube-apiserver
```
tfussell, line 377: This image is only used to be notified when there's a new version and to extract the version by running `kube-apiserver --version` in the `Dockerfile`. It's not pretty, but it works.

images.yaml
```diff
@@ -374,6 +374,24 @@
...
+  tagTrimVersionPrefix: true
+  overrideRepoName: hyperkube
```
tfussell, line 379: Push to `quay.io/giantswarm/hyperkube`.

images.yaml
```diff
@@ -374,6 +374,24 @@
     - tagSuffix: giantswarm
       dockerfileOptions:
       - RUN apt-get update && apt-get install -y ca-certificates && rm -rf /var/lib/apt/lists/*
+- name: k8s.gcr.io/kube-apiserver
+  tagTrimVersionPrefix: true
```
tfussell, line 378: Added to transform `v1.20.0` into `1.20.0` to match https://github.com/giantswarm/k8scloudconfig/blob/cbfb99e4baec60525469f8b262fffc86dddad7e4/pkg/template/release.go#L19. Alternatively, we would need a new major release of `k8scloudconfig` that doesn't strip the "v" prefix I think.

pkg/registry/registry_docker.go
```diff
@@ -169,7 +169,7 @@ func (r *Registry) logDocker(reader io.Reader) error {
 			return microerror.Maskf(dockerError, "docker task failed: %s", logMsg)
 		}
 
-		r.logger.Log("level", "debug", "message", fmt.Sprintf("docker status"), "docker", string(s.Bytes()))
+		r.logger.Log("level", "debug", "message", "docker status", "docker", string(s.Bytes()))
```
tfussell, line 172: Unrelated boyscouting.

stone-z (approved):
Cool idea!

pkg/retagger/job_definition.go
```diff
@@ -160,10 +160,10 @@ func fromImageTagPattern(image images.Image, tagPattern images.TagPattern) (JobD
 	j := JobDefinition{
 		SourceImage:   image.Name,
 		SourcePattern: tagPattern.Pattern,
-	}
-
-	if image.OverrideRepoName != "" {
-		j.Options.OverrideRepoName = image.OverrideRepoName
+		Options: JobOptions{
+			TagTrimVersionPrefix: image.TagTrimVersionPrefix,
+			OverrideRepoName:     image.OverrideRepoName,
```
MarcelMue, line 165: Is this fine if we don't check for the empty case first? I am not sure how the code is handled.

MarcelMue (approved):
Generally LGTM

pkg/retagger/job_definition.go
tfussell: By removing the check, when `OverrideRepoName` is the empty string, we would assign an empty string to `OverrideRepoName` which is already initialized to the zero value (no pointers involved), so I don't see how it would affect the behavior.

PR #580, opened 2021-06-14
Force update for alpine:3.13

marians:


images.yaml
```diff
@@ -1,4 +1,7 @@
 - name: alpine
+  tags:
+  - sha: 69e70a79f2d41ab5d637de98c1e0b055206ba40a8145e7bddb55ccc04e13cf8f
+    tag: 3.13
```
marians, line 4: This is the current SHA for `3.13` and `3.13.5`

stone-z: I think we should tag it as 3.13.5, not sure if retagger follows minor aliases like that
```suggestion
    tag: 3.13.5
```

marians: My aim here was to update what we have in quay tagged as 3.13.

stone-z: :+1: I see, it looks like it indeed does not follow.

stone-z (approved):

PR #599, opened 2021-07-27
Retag sonobuoy on gs alpine

stone-z:


stone-z:
Tried this out using https://github.com/giantswarm/sonobuoy-alpine/blob/9f1ef327f35d0b5cc5b06aae15783a7b93629075/Dockerfile. Will try the image retagger produces to make sure it also works

stone-z:
Going with a ping because just retagging

PR #606, opened 2021-08-06
Exclude quay.io/calico/kube-controllers v3.19.2 as it's breaking the execution

marians:
See https://gigantic.slack.com/archives/C3C7ZQXC1/p1628262522004900 for internal discussion

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -637,10 +637,8 @@
 - name: quay.io/calico/kube-controllers
   overrideRepoName: calico-crd-installer
   patterns:
-  # v3.19.2 is not functional, so we treat is "as is" without
-  # applying our additional build options.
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -637,7 +637,8 @@
 - name: quay.io/calico/kube-controllers
   overrideRepoName: calico-crd-installer
   patterns:
-  - pattern: '>= v3.15.0'
+  # Skip v3.19.2 as it's not functional
+  - pattern: '>= v3.15.0 || != v3.19.2'
```
ericgraf, line 641: Should this be an and not a or

ericgraf: '>= v3.15.0 != v3.19.2'

ericgraf (approved):

PR #615, opened 2021-08-24
Add calicoctl

tfussell:
Adding calicoctl for https://github.com/giantswarm/giantswarm/issues/14791. Using alpine as the base image so that we can use the shell (upstream image is FROM scratch).

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -641,6 +641,16 @@
 - name: quay.io/calico/cni
   patterns:
   - pattern: '>= v3.9.5'
+- name: quay.io/calico/ctl
+  patterns:
+  - pattern: '>= v3.19.0'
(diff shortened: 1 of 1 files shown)


MarcelMue (approved):
Looks reasonable to me

images.yaml
```diff
@@ -641,6 +641,16 @@
 - name: quay.io/calico/cni
   patterns:
   - pattern: '>= v3.9.5'
+- name: quay.io/calico/ctl
```
stone-z, line 644: Maybe override the repo name, since this will just make it `giantswarm/ctl`

tfussell: Oops, yes, I meant to do that. Thanks.

PR #622, opened 2021-09-08
Add starboard related images

ericgraf:
Add images that starboard references

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -56,12 +56,12 @@
 - name: aquasec/kube-bench
   patterns:
   - pattern: '>= 0.2'
-- name: aquasec/starboard-operator
-  patterns:
-  - pattern: '>= 0.11.0'
 - name: aquasec/scanner
   patterns:
   - pattern: '>= 5.3'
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -613,6 +622,9 @@
 - name: omnition/opencensus-collector
   patterns:
   - pattern: '>= 0.1.11'
+- name: openpolicyagent/conftest
+  patterns:
+  - pattern: '>= v0.25.0'
```
MarcelMue, line 627: 0.27.0 exist AFAICT

ericgraf: Picked 0.25.0 because it's specified here for starboard v0.11.0  https://github.com/aquasecurity/starboard/blob/v0.11.0/deploy/helm/values.yaml#L304

MarcelMue: Maybe we can add a PR to update this for them? Like even on master they are using outdated conftest :D

MarcelMue (approved):
Ok with me for now

PR #628, opened 2021-09-14
Add dockerhub to push registry in ci job

ericgraf:
Add dockerhub as push registry

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,6 +81,46 @@ jobs:
           name: Execute retagger
           command: ./retagger -r ${REGISTRY} -o ${REGISTRY_ORGANISATION} -u ${REGISTRY_USERNAME} -p ${REGISTRY_PASSWORD}
 
+
+  e2eDockerhub:
+    environment:
+      REGISTRY: "docker.io"
+      REGISTRY_ORGANISATION: "giantswarm"
+    machine: true
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Set registry credentials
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -81,6 +81,46 @@ jobs:
...
+          command: |
+            echo 'export REGISTRY_USERNAME=${DOCKERHUB_USERNAME}' >> $BASH_ENV
+            echo 'export REGISTRY_PASSWORD=${DOCKERHUB_PASSWORD}' >> $BASH_ENV
+            source $BASH_ENV
+      - run:
+          name: Run e2e test
+          command: |
+            ./e2e.test -test.run ".*E2e.*" -test.v
+
+  retagDockerhub:
```
MarcelMue, line 104: Should we just drop quay completely? Or do we expect issues with that?

.circleci/config.yml
```diff
@@ -81,6 +81,46 @@ jobs:
           name: Execute retagger
           command: ./retagger -r ${REGISTRY} -o ${REGISTRY_ORGANISATION} -u ${REGISTRY_USERNAME} -p ${REGISTRY_PASSWORD}
 
+
+  e2eDockerhub:
```
MarcelMue, line 85: This seems to be failing, but not entirely sure why.

.circleci/config.yml
stone-z: I think a bunch of places are still hardcoded to use quay, so should probably keep it until there's been an organized move away

PR #649, opened 2021-11-01
Capo images

ericgraf:


images.yaml
```diff
@@ -516,6 +516,15 @@
 - name: justwatch/elasticsearch_exporter
   patterns:
   - pattern: '>= v1.1.0'
+- name: k8scloudprovider/cinder-csi-plugin
+  patterns:
+  - pattern: '>= 1.22.0'
```
tfussell, line 521: Are these tied to specific Kubernetes versions? If so are we aiming to support k8s 1.22+ only?

ericgraf: Thanks good catch. I had just picked latest.

MarcelMue (approved):

pipo02mix (approved):

PR #650, opened 2021-11-02
Bumping images for Flux

ljakimczuk:


images.yaml
```diff
@@ -271,6 +271,8 @@
     tag: v0.14.1
   - sha: e5dcbe48c1d032e0103997de4551df8f6576f9a9e91e8e3a768410171c2fc5b8
     tag: v0.15.0
+  - sha: 4140e621065c5ec7d98052dab50f3cd32e59e75e135c7e7845103fc482cf5809
```
MarcelMue, line 274: Should we just change this to a regex which is `>v0.15.0` like we do for other projects?

ljakimczuk: I guess that would be simpler, but maybe there was some specific reason to list them by tags and shas.

rossf7: Originally we only supported tags and SHAs The pattern support was added later. I'm not sure why we didn't use it when setting up Flux 

I agree with @MarcelMue it would be useful here. We want to keep in sync with FluxCD releases so this will reduce some toil and I don't see a downside.

MarcelMue (approved):

images.yaml
ljakimczuk: Got it, thank you @MarcelMue @rossf7

PR #658, opened 2021-12-21
Fixing repo name

ljakimczuk:
Seems we have already hosted `yamllint` repo in quay, so changing retagged one's name.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -148,6 +148,7 @@
   patterns:
   - pattern: '>= 7.67.0'
 - name: cytopia/yamllint
+  overrideRepoName: cytopia-yamllint
   patterns:
   - pattern: '>= 1.9'
 - name: dduportal/bats


MarcelMue (approved):

rossf7:
> Seems we have already hosted yamllint repo in quay, 

Looks like the repo was created today. IIRC retagger can create the repo?
https://quay.io/repository/giantswarm/yamllint

ljakimczuk:
> Looks like the repo was created today. IIRC retagger can create the repo?

Hmm, you are right, I've misinterpreted it, I thought this repo was already there, but maybe that's not the case and instead it got created private today

rossf7:
@ljakimczuk Can you switch the repo to be public?

Then I think we can close this. But the docs are outdated here https://github.com/giantswarm/retagger#background-and-details

ljakimczuk:
@rossf7 sure, I'm closing this one. I will try to take care of the docs as well, but first I would need to make sure that it was the retagger who created this quay repository.. because right now I'm not sure, because we do seem to have some old yamllint related GH repos: https://github.com/giantswarm/yamllint-docker. But it's quite old, so hard to say if quay repo was originally created back then or not

PR #693, opened 2022-05-11
Keep retagging even when one image fails

QuentinBisson:
To avoid failing all images when the changes are not related, we retag all images even if one fail, then error out. https://gigantic.slack.com/archives/C3C7ZQXC1/p1652278502391569


QuentinBisson:
@giantswarm/team-honeybadger could I get your help to test this ?

PR #699, opened 2022-06-09
Retag trivy-operator

stone-z:
Aqua is going away from starboard in favor of trivy-operator. We will need to follow it


stone-z:
CI failure seems unrelated. Has been flaky in the past

fhielpos (approved):
LGTM, if IRC that check always failed

PR #723, opened 2022-08-24
Do not copy /lib in calico-crd-installer

ubergesundheit:
This PR attempts to fix failing retagger runs since calico released https://github.com/projectcalico/calico/releases/tag/v3.24.0 which contains this change https://github.com/projectcalico/calico/pull/6225 which removed copying from `/lib`

I did the same and the image builds correctly on my machine.


ubergesundheit:
ping @giantswarm/team-rocket I think you're the owner here. merging now

PR #748, opened 2022-10-26
Re enable kyverno image

fhielpos:
Re enabling Kyverno images pointing to specific SHA since now Kyverno uses OCI index instead.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -396,9 +396,13 @@
     tag: v3.9-thick-amd64
   - sha: ce4c3a421b0ca97540af911f34858bdd9b06fa5817ba7fa2b3bbb906f9237b46
     tag: v3.9-amd64
-# - name: ghcr.io/kyverno/kyverno
-#   patterns:
-#   - pattern: '>= v1.3.4'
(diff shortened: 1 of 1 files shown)


images.yaml
```diff
@@ -396,9 +396,13 @@
...
-#   patterns:
-#   - pattern: '>= v1.3.4'
+- name: ghcr.io/kyverno/kyverno
+  tags:
+  - sha: 86577d2e18c1b23b0cc9d76f1f2ee49fea429d4fa7dfbb91585e53ffb5b32dcb
+    tag: v1.8.0
+  - sha: 14c019a58a27355daea7b5e354a5f90e31e6affa18312dccbfc61d4779eb4591
+    tag: v1.8.1
+# Momentarily disable kyvernopre
 # - name: ghcr.io/kyverno/kyvernopre
```
stone-z, line 406: Need kyvernopre too

fhielpos: I know, wanted to check if this worked since I have tested it only with the arm SHA

fhielpos: NVM, let's just yolo and test it all together

PR #774, opened 2022-12-15
Added kubeclarity images

Strigix:
Could someone please review whether I've added these images correctly and if so accept the PR? 
Repositories are already created on quay.


Strigix:
The checks fail on different images than those added?

uvegla:
FTR Needs to wait for Aliyun limit increase.

Strigix:
This is no longer needed since we decided not to go through with using kubeclarity for the time being!

PR #780, opened 2023-01-11
Do not retag if there are two jobs writing different content to the same `repo.hostname/image:tag` destination, add coredns/etcd images usable by kubeadm/CAPI clusters

AndiDog:
The conflict check is for my paranoia. We need to ensure that for example the 2 different etcd images (which actually have different content) never get confused. My config comments explain the naming schemes, use case and why we don't expect conflicts.

This change allows kubeadm to pull official Kubernetes images from `docker.io/giantswarm/...` (https://github.com/giantswarm/roadmap/issues/1722), as done in CAPI clusters (and generally all other standard kubeadm-administered clusters).

Other than the 2 added image configs, these is the full list and I found the others to be correctly configured in retagger, and pullable, already:

```
λ kubeadm config images list --kubernetes-version v1.23.15 --image-repository docker.io/giantswarm 2>/dev/null
docker.io/giantswarm/kube-apiserver:v1.23.15
docker.io/giantswarm/kube-controller-manager:v1.23.15
docker.io/giantswarm/kube-scheduler:v1.23.15
docker.io/giantswarm/kube-proxy:v1.23.15
docker.io/giantswarm/pause:3.9
docker.io/giantswarm/etcd:3.5.6-0
docker.io/giantswarm/coredns:v1.10.0
```

To reviewers: the E2E test fails with `Image retagger-e2e/2.6 does not exist.` even though the test file says it should (`only 2.6 exists`). Please help me on that one.

### DIFF
--- a/images.yaml
+++ b/images.yaml
@@ -1097,6 +1097,20 @@
 - name: registry.k8s.io/cluster-proportional-autoscaler-amd64
   patterns:
   - pattern: '>= 1.6.0'
+
+# We have to ensure that etcd is available to CAPI clusters - we set
+# `clusterConfiguration.imageRepository=docker.io/giantswarm` (e.g. in cluster-aws:
+# https://github.com/giantswarm/cluster-aws/search?q=imageRepository%3A) and that means `kubeadm join` will use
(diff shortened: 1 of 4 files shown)


AverageMarcus (approved):
LGTM!

PR #789, opened 2023-01-26
Revert "remove kyverno policy reporter"

ubergesundheit:
Reverts giantswarm/retagger#788

@giantswarm/team-shield fyi


fhielpos:
Seems that these images moved to OCI. Will need to use the `sha` directly. Will update this PR later today.

fhielpos:
Not needed anymore 😄

Issue #793, opened 2023-02-01
Add support for Harbor v2 API


danibaeyens:
retagger works fine for quay.io and alibaba's public registries, but there are companies using their own registries with Harbor which can benefit a lot from retagger. It would be cool supporting harbor

PR #794, opened 2023-02-01
Adds Harbor registry V2 API support to retag images to a Harbor destination

danibaeyens:
Adds an optional flag to select the registry type to be used, using quay.io as a default.
Adds Harbor registry V2 API support to retag images to a Harbor destination


danibaeyens:
I've found that the current logic is too tight to the original quay supported implementation. I wanted to avoid making a whole refactor to correctly use the different types, so, I've just added a similar hook to the existent one for alibaba's. The biggest difference is that for not hardcoding any specific harbor registry (as they are private instances), I've added a flag which helps to select the registry to be used

Fixes #793

Gacko:
@giantswarm/team-honeybadger Can you please verify if this is still a thing today and close the PR if not? 🙂

uvegla:
I dont think its relevant anymore, closed it, thanks! 👍🏻

PR #816, opened 2023-03-28
Redundant image

jkremser:
In my previous PR (https://github.com/giantswarm/retagger/pull/814) I've added an image that has already been retagged&synced. So I am removing the redundancy here.

Also adding a note to the readme that those repositories need to be created. Otherwise sync fails.

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,7 +258,8 @@ images.
 
 ## How to add your image to the job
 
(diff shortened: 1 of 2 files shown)


kubasobon (approved):
Yeah, good catch. quay.io definitely requires the repo to be pre-created.

PR #822, opened 2023-04-18
Add prometheus stringlabels

QuentinBisson:
Towards https://github.com/giantswarm/giantswarm/issues/26238

### DIFF
--- a/images/skopeo-docker-io.yaml
+++ b/images/skopeo-docker-io.yaml
@@ -99,10 +99,8 @@ docker.io:
     prom/cloudwatch-exporter:
       - "cloudwatch_exporter-0.4"
     prom/prometheus:
-      - "v2.6.1"
-      - "v2.9.1"
-      - "v2.13.1"
-      - "v2.15.2"
+      - ">= 2.30.0"
+      - ">= 2.43.0-stringlabels"
     python:
       - "3.8.12-slim"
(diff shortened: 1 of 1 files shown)


images/skopeo-docker-io.yaml
```diff
@@ -99,10 +99,8 @@ docker.io:
     prom/cloudwatch-exporter:
       - "cloudwatch_exporter-0.4"
     prom/prometheus:
-      - "v2.6.1"
-      - "v2.9.1"
-      - "v2.13.1"
-      - "v2.15.2"
+      - ">= 2.30.0"
+      - ">= 2.43.0-stringlabels"
```
hervenicol, line 103: IIRC, the ">=" operator splits the version number, so it may require a "simple" version and may not work with `-stringlabels`

hervenicol: Oh, forget about it. We've moved to skopeo since then.

hervenicol (approved):
The version+suffix syntax works for memcached (https://github.com/giantswarm/retagger/blob/main/images/skopeo-docker-io.yaml#L210), so LGTM

hervenicol:
Oh, it may conflict with this one:
https://github.com/giantswarm/retagger/blob/main/images/skopeo-docker-io.yaml#L223

PR #863, opened 2023-08-29
Add special handling of known mutable tags

AverageMarcus:
To ensure that known mutable tags are synced from upstream we add a special handling for `latest`, `debug` and `develop` to always attempt to sync changes from upstream.


kubasobon (approved):
:+1:

PR #864, opened 2023-08-29
Added Sonobuoy latest

AverageMarcus:
Needed for the conformance tests pipeline to avoid the Docker Hub rate limiting. 

Towards: https://github.com/giantswarm/giantswarm/issues/28015

### DIFF
--- a/images/skopeo-docker-io.yaml
+++ b/images/skopeo-docker-io.yaml
@@ -101,6 +101,8 @@ docker.io:
       - "4.0.9"
       - "6.2.1-alpine"
       - "6.2.4-alpine"
+    sonobuoy/sonobuoy:
+      - "latest"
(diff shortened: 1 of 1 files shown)


images/skopeo-docker-io.yaml
```diff
@@ -210,6 +210,8 @@ docker.io:
       - ">= 2.43.0-stringlabels"
     prom/pushgateway:
       - ">= v1.0.0"
+    sonobuoy/sonobuoy:
+      - "latest"
```
kubasobon, line 214: This is not a semver rule and should go into `docker.io > images`, not `docker.io > images-by-semver`

AverageMarcus: Thanks! Updated!

PR #880, opened 2023-09-25
Images: Add `ingress-nginx/controller-chroot`.

Gacko:


ced0ps (changes requested):
Is there a reason you don't do a pattern match like 
```
- image: registry.k8s.io/etcd
  semver: ">= v3.5.4-0"
  add_tag_suffix: k8s
  ```

Gacko:
Because we are matching on the SHA hash of the source repository, too.

ced0ps (approved):
LGTM

PR #885, opened 2023-10-09
Update Squid proxy image to allow SSH proxying

AverageMarcus:
Towards: https://github.com/giantswarm/giantswarm/issues/26154

This PR enables support for proxying ssh over VPN via the squid. 

@calvix was there a reason for using the specific sha? According to Docker Hub the current sha for `5.2-22.04_beta` is `b37e34479db87f2f561e89747d09db69899b485fb549892acd3af7485e3ba102`


AverageMarcus:
Turns out Packer only supports proxying ssh over SOCKS5 so this wouldn't help us. See https://github.com/giantswarm/giantswarm/issues/26154#issuecomment-1752754665

PR #886, opened 2023-10-09
Added go-socks5-proxy

AverageMarcus:
Towards: https://github.com/giantswarm/giantswarm/issues/26154

@erkanerol I went with the same SHA you did in your tests. Was there anything specific about that or was it just the latest at the time?


erkanerol:
It was just the latest :)

PR #932, opened 2023-11-13
Disable Renovate

marians:
CI runs for retagger are quite expensive due to layer caching and due to the long duration of job runs.

To limit our cost, we are disabling dependency updates by Renovate.


marians:
The failed CircleCI test is not related to the change.

PR #933, opened 2023-11-13
Phoenix retagger update v3 - grand finale

T-Kukawka:
- pining gcp compute to latest used
- kiam - oldest release uses 4.2.0
- removing ciprod - archived
- Vault - latest release from April uses 1.5.4 - can we use that one @whites11 ?

### DIFF
--- a/images/skopeo-gcr-io.yaml
+++ b/images/skopeo-gcr-io.yaml
@@ -17,6 +17,6 @@ gcr.io:
     cadvisor/cadvisor:
       - ">= v0.44.0"
     k8s-staging-cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:
-      - "v1.8.0"
+      - "1.8.0"
     kubebuilder/kube-rbac-proxy:
       - ">= v0.4.1"
--- a/images/skopeo-gcr-io.yaml
+++ b/images/skopeo-gcr-io.yaml
@@ -17,6 +17,6 @@ gcr.io:
     cadvisor/cadvisor:
       - ">= v0.44.0"
     k8s-staging-cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:
-      - ">= 1.7.0"
+      - "v1.8.0"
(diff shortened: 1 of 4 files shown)


fiunchinho (approved):

whites11:
on vintage vault vms we use 1.7.10 https://github.com/giantswarm/hive/blob/3183439a7c7099dc572c69bf1eac545795218d0c/vars.yml#L14

on gorilla/rfjh2 we use 1.5.4:

```
$ kubectl -n vault-app get po vault-5d4bb89fb8-wpz4x -o yaml | grep image:
    image: quay.io/giantswarm/vault:1.5.4
```

so I guess >= 1.5.4 should be ok

images/skopeo-gcr-io.yaml
```diff
@@ -17,6 +17,6 @@ gcr.io:
     cadvisor/cadvisor:
       - ">= v0.44.0"
     k8s-staging-cloud-provider-gcp/gcp-compute-persistent-disk-csi-driver:
-      - ">= 1.7.0"
+      - "v1.8.0"
```
whites11, line 20: did you want a >= and no `v` here?

T-Kukawka: i want 1.8.0 only, my bad

T-Kukawka: fixed

whites11: is it like that somewhere else? you might still need the `=` operator

whites11 (approved):

images/skopeo-gcr-io.yaml
T-Kukawka: yeah there are other examples: https://github.com/giantswarm/retagger/blob/7ed2f2593bc63a5c264bfddbdb7afa384c7d56e1/images/skopeo-gcr-io.yaml#L5

whites11: perfect LGTM then

PR #938, opened 2023-11-16
Updated Falco images

Strigix:


marians (approved):

stone-z (approved):
Here or in a followup we should also ditch `prowler`, `polaris`, `kube-bench`, and check out `gitleaks`. I bet we can dump OPA gatekeeper now too.

Strigix:
I'll make a separate one for those.

PR #939, opened 2023-11-17
Add gsoci.azurecr.io as push target

kubasobon:
Towards https://github.com/giantswarm/roadmap/issues/2895

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,8 +20,8 @@ import (
 )
 
 var (
-	temporaryWorkingDir     = path.Join(os.TempDir(), "retagger")
 	skopeoSyncOutputPattern = regexp.MustCompile(`Would have copied image.*?from="docker://(.*?)[@:](.*?)".*`)
+	temporaryWorkingDir     = path.Join(os.TempDir(), "retagger")
 
 	flagLogLevel         string
 	flagExecutorCount    int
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -33,9 +33,11 @@ const (
 	customizedImagesFile = "images/customized-images.yaml"
 	defaultPlatform      = "linux/amd64"
 	dockerTransport      = "docker://"
-	quayURL              = "quay.io/giantswarm"
-	aliyunURL            = "giantswarm-registry.cn-shanghai.cr.aliyuncs.com/giantswarm"
 	filteredFileSuffix   = ".filtered"
+
+	aliyunURL = "giantswarm-registry.cn-shanghai.cr.aliyuncs.com/giantswarm"
+	azureURL  = "gsoci.azurecr.io"
+	quayURL   = "quay.io/giantswarm"
```
piontec, line 40: for the time being, we also need to push to the old ACR, `giantswarm.azurecr.io`

piontec: Can we also make this configurable? We will have to change it a few times in the coming months, recompiling each time doesn't seem very appealing...

main.go
```diff
@@ -119,10 +122,13 @@ func (img *CustomImage) RetagUsingSHA() error {
 	if len(img.DockerfileExtras) == 0 {
 		source := fmt.Sprintf("%s%s@sha256:%s", dockerTransport, img.Image, img.SHA)
 		wg := sync.WaitGroup{}
-		wg.Add(2)
+		wg.Add(3)
 		// Quay
 		destination := fmt.Sprintf("%s%s/%s:%s", dockerTransport, quayURL, destinationName, destinationTag)
 		go copyImage(&wg, errorCounter, source, destination)
+		// AzureCR
```
piontec, line 129: Shouldn't we use `oci://` transport now?

kubasobon: I don't think so, no.

main.go
kubasobon: Sure, I can make it more configurable.

kubasobon: Actually, it would require similar effort. Let's talk offline

main.go
```diff
@@ -33,9 +33,11 @@ const (
 	customizedImagesFile = "images/customized-images.yaml"
 	defaultPlatform      = "linux/amd64"
 	dockerTransport      = "docker://"
-	quayURL              = "quay.io/giantswarm"
-	aliyunURL            = "giantswarm-registry.cn-shanghai.cr.aliyuncs.com/giantswarm"
 	filteredFileSuffix   = ".filtered"
+
+	aliyunURL = "giantswarm-registry.cn-shanghai.cr.aliyuncs.com/giantswarm"
+	azureURL  = "gsoci.azurecr.io"
```
marians, line 39: Should there be the `/giantswarm` suffix after `gsoci.azurecr.io`?

kubasobon: It should! Thank you!

PR #946, opened 2023-12-01
new image for cluster-api-ipam-provider-in-cluster 

jkremser:
weird version format so the `>=` wouldn't probably match this


glitchcrab (approved):

jkremser:
let's not merge yet actually, it looks like the image is missing because their pipeline is broken - https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/actions/runs/7060469599/job/19220064015 I will try to help them to fix it

glitchcrab:
i've assigned myself to the PR so that it doesn't get lost when Jiri leaves

glitchcrab:
i'm going to go ahead and merge this in

PR #948, opened 2023-12-07
Add new hugo images

MarcelMue:


MarcelMue:
I hope this is the correct way to update images currently, but didn't find any other info.

PR #951, opened 2023-12-11
Enable golang based on alpine 3.19

marians:


marians:
@kubasobon Any idea why this failed?

marians:
Obsolete because of https://github.com/giantswarm/retagger/pull/958

PR #954, opened 2023-12-19
add gravitational/teleport

ssyno:


anvddriesch (approved):

tuladhar:
@ssyno We might also need to re-tag `gravitational/teleport-distroless`

As mentioned in [teleport-kube-agent chart](https://github.com/giantswarm/teleport-kube-agent-app/blob/df393fd33fe5ac6b245dcecc60fc0bf8092ae605/helm/teleport-kube-agent/values.yaml#L212C1-L218C63) 
```
# Since version 13, hardened distroless images are used by default.
# You can use the deprecated debian-based images by setting the value to
# `public.ecr.aws/gravitational/teleport`. Those images will be
# removed with teleport 14.
image:
  repository: public.ecr.aws/gravitational/teleport-distroless
```

PR #955, opened 2023-12-20
limit runs to main branch

kubasobon:
Limits CircleCI jobs to `validate-images-yaml` on branches other than `main`.


piontec (approved):
LGTM

PR #960, opened 2024-01-19
Add giantswarm/staticjscms-hugo-standalone

lyind:
Towards https://github.com/giantswarm/giantswarm/issues/29451

### Summary

We can't currently make `ghcr.io` images (GitHub calls them packages) public, because it's disabled in organization settings (this seems to make sense).

However, I need an image to be publicly accessible for devs local machines (ease of use).

This PR adds it to retagger.

### Note

Changing the workflow to push to quay.io/docker.io directly may happen later.

### DIFF
--- a/images/skopeo-ghcr-io.yaml
+++ b/images/skopeo-ghcr-io.yaml
@@ -15,6 +15,8 @@ ghcr.io:
       - ">= 0.16.1"
     external-secrets/external-secrets:
       - ">= v0.7.0"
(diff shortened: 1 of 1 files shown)


marians (approved):
Not against this, but if we are the creators of the image, why go through retagger to get it into our registries? Why not build it via architect-orb `push-to-registries` in CircleCI?

marians:
This makes sense. I just hadn't seen this case yet.

PR #972, opened 2024-02-09
Add validation for skopeo files

marians:
Context: https://github.com/giantswarm/giantswarm/issues/28543

So far we don't validate skopeo YAML files in CI. This allows to introduce schema errors without noticing in a PR.

This PR adds a `skopeo sync --dry-run` call for each YAML file starting with `skopeo-`.


mproffitt (approved):
LGTM

PR #975, opened 2024-02-21
Update skopeo-eu-gcr-io.yaml

njuettner:
Towards: https://github.com/giantswarm/roadmap/issues/2924


weseven (approved):

marians:
@njuettner Is the goal here to pick up new versions automatically in the future? In that case it would be possible to specify a semver range, but under `images-by-semver` instead of `images`.

njuettner:
@marians there's now a different source where we get newer images, I only removed the older ones which we definitely don't use anymore.

PR #978, opened 2024-02-23
correct ipam-in-cluster tag

glitchcrab:
following on from #946, it seems that the alpha tags no longer exist so we need to use a newer image.

available tags are listed here https://console.cloud.google.com/gcr/images/k8s-staging-capi-ipam-ic/GLOBAL/cluster-api-ipam-provider-in-cluster?pli=1


vxav (commented):
Why no [0.1.0](https://github.com/kubernetes-sigs/cluster-api-ipam-provider-in-cluster/releases/tag/v0.1.0) ?

vxav (approved):

piontec (approved):

PR #996, opened 2024-04-04
Make static skopeo binary available in our container image

marians:
Previously we tried copying the skopeo binary from a container. However, since it is dynamically linked against system libraries, and these libraries were not portable, this approach failed.

[Error](https://app.circleci.com/pipelines/github/giantswarm/retagger/3412/workflows/d5fa7b11-ba82-45f6-a121-102c9da9cef7/jobs/67936):

```
skopeo: error while loading shared libraries: libgpgme.so.11: cannot open shared object file: No such file or directory
```

Unfortunately, skopeo does not provide binaries as part of their releases. They provide an alpine package, but it's lagging behind. We need v1.15, but the Alpine package is at v1.14.

With this PR we are building our own static skopeo binary to have it available in the same container image that provides retagger.

This binary does not provide GPG functions, which (as I assume) we don't need anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,21 +47,6 @@ jobs:
                 grep "level=error" dry-run.log
                 exit 1
             fi
-            
-
-  build-go:
-    machine:
-      image: ubuntu-2204:2024.01.2
-    steps:
-      - checkout
-      - run:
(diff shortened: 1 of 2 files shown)


allanger (approved):
lgtm

PR #1012, opened 2024-05-21
Add latest version hugo

pipo02mix:


QuentinBisson (approved):
Do we still need thé old ones?

pipo02mix:
I am not sure if other project uses it, but I have in my todo list the check

PR #1018, opened 2024-06-13
Fix retry errors

marians:
We see different shell scripting errors resulting in retries for `docker login` not working.

The errors indicate that, contrary to our expectations, the step commands are executed using `/bin/sh` instead of bash.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -265,7 +265,6 @@ build_and_retag: &build_and_retag
           branches:
             only:
               - main
-              - fix-retries # TODO: remove before merging
         requires:
           - build-and-push-docker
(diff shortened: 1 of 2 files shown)


marians:
This just ran and all jobs succeeded. See https://app.circleci.com/pipelines/github/giantswarm/retagger/3550

I will revert the branch trigger with the next commit.

mproffitt (approved):

marians:
Here we see how the retry actually does its jobs:

image

https://app.circleci.com/pipelines/github/giantswarm/retagger/3550/workflows/071be565-abc4-46bf-831a-976796f543e5/jobs/72218

PR #1037, opened 2024-08-19
Add opensearch and traefik

marians:
Towards https://github.com/giantswarm/giantswarm/issues/26502

This adds traefik, which I intend to use as a reverse proxy for our docs and search setup, and OpenSearch, which will be our search engine going forward.


pipo02mix (approved):

marians:
No, currently not planned

PR #1044, opened 2024-10-07
Update dex image source

QuentinBisson:
Dex images comes from ghcr and not quay


fhielpos (approved):

fhielpos:
I think this should be `ghcr` and not `gcr`:
https://github.com/dexidp/dex/pkgs/container/dex

QuentinBisson:
And you're right 🤦🏻

PR #1053, opened 2024-10-30
Add custom job for Trivy databases retag

fhielpos:
Towards: https://github.com/giantswarm/giantswarm/issues/31965

This is an experimental PR to add the Trivy databases to our registries and keeping them in sync


.circleci/config.yml
```diff
@@ -490,3 +586,13 @@ workflows:
...
     <<: *build_and_retag
+
+  # Runs every 12 hours at 09:08 and 21:08 UTC
+  trivy_databases:
+    triggers:
+      - schedule:
+          cron: "08 09,21 * * *"
+          filters:
+            branches:
+              only: main
```
fhielpos, line 597: Maybe this is not relevant to have, but better be safe than sorry

.circleci/config.yml
```diff
@@ -129,6 +129,55 @@ jobs:
             - "<<parameters.filepath>>.filtered"
             - "<<parameters.filepath>>.filtered.errlog"
 
+  retag-trivy-dbs:
```
fhielpos, line 132: This can be re-used for any image, maybe it could have another name

fhielpos:
Since now the job is a cronjob, it won't run on push to this branch. Here is the last workflow run:
https://app.circleci.com/pipelines/github/giantswarm/retagger/3787/workflows/b7b2a8da-4929-4ad5-8461-59f9b0c13fe2

.circleci/config.yml
stone-z: Why not just call `retag-registry` with a yaml that contains the DB repositories?

fhielpos: I tried it, but it requires way too much work to validate the files and it's too prone to fail, so I reverted to this approach which runs faster and doesn't require as many moving pieces.

stone-z (approved):

PR #1072, opened 2025-01-21
Images: Retag `ahrtr/etcd-defrag`.

Gacko:


images/skopeo-ghcr-io.yaml
```diff
@@ -1,6 +1,7 @@
 # Docs: https://github.com/kubasobon/skopeo/blob/main/docs/skopeo-sync.1.md#yaml-file-content-used-source-for---src-yaml
 ghcr.io:
   images:
+    ahrtr/etcd-defrag: []
```
AverageMarcus, line 4: Is this valid? Why not use the `images-by-semver`?

Gacko: https://github.com/kubasobon/skopeo/blob/main/docs/skopeo-sync.1.md#yaml-file-content-used-source-for---src-yaml

> Repository `registry.example.com/busybox`: all images, as no tags are specified.

Gacko: So, yes, I assume this is valid. Or would you really just retag images matching a semver?

AverageMarcus: I mean, do we need the older images? I assume you're currently wanting to use v0.22.0? Lets just retag that and newer?

Gacko: Yeah, makes sense. Also my current approach would probably also retag untagged images or at least development ones.

Gacko: Fixed!

AverageMarcus (approved):

PR #1087, opened 2025-02-25
add kube-webhook-certgen v1.5.1

hervenicol:
This image is required for latest kube-prometheus-stack (https://github.com/giantswarm/kube-prometheus-stack-app/pull/443).

I did it just like the previous versions were done, but maybe there's a better way so we always have latest updates?

### DIFF
--- a/images/renamed-images.yaml
+++ b/images/renamed-images.yaml
@@ -528,6 +528,10 @@
   tag_or_pattern: "v1.5.0"
   sha: aaafd456bda110628b2d4ca6296f38731a3aaf0bf7581efae824a41c770a8fc4
   override_repo_name: ingress-nginx-kube-webhook-certgen
+- image: registry.k8s.io/ingress-nginx/kube-webhook-certgen
+  tag_or_pattern: "v1.5.1"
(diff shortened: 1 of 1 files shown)


Gacko:
These images are being referenced by their digest, because we at some point had trouble using the digest in our Ingress NGINX chart. So we at least wanted to make sure they are synced by their hash.

Gacko:
Ah and as I am usually the person releasing them upstream and taking care to have our Ingress NGINX chart synced to upstream, I'm also always adding the latest releases here.

What makes me wonder: IIRC We are not referencing this tag in the latest release of Ingress NGINX, so Prometheus must be watching these releases somehow else. 😅

hervenicol:
> What makes me wonder: IIRC We are not referencing this tag in the latest release of Ingress NGINX, so Prometheus must be watching these releases somehow else. 😅

It seems they get the latest tag from here: https://github.com/kubernetes/ingress-nginx/blob/main/images/kube-webhook-certgen/TAG

Gacko:
Hopefully not... this tag does not represent what's really being promoted. Anyway, if it works for you, I'm fine.

PR #1097, opened 2025-04-15
Add new kyverno-plugin image

fhielpos:


images/skopeo-ghcr-io.yaml
```diff
@@ -32,7 +32,7 @@ ghcr.io:
     kyverno/kyvernopre: ">= v1.9.0"
     kyverno/kyverno-cli: ">= v1.12.5"
     kyverno/policy-reporter: ">= 2.12.0"
-    kyverno/policy-reporter-kyverno-plugin: ">= 1.5.0"
+    # kyverno/policy-reporter-kyverno-plugin: ">= 1.5.0"
```
stone-z, line 35: Just curious - what was this 1.5.0 version number actually for if not the plugin?

stone-z (approved):

images/skopeo-ghcr-io.yaml
fhielpos: It was for the plugin, but they reworked the whole system and moved to a monorepo so that image got deprecated.

PR #1106, opened 2025-06-27
Add peaceiris/hugo:v0.146.4-full to skopeo-docker-io.yaml

lyind:
Towards https://github.com/giantswarm/giantswarm/issues/33610


marians (approved):
FYI You may delete all previous versions, as they have been synced already to all regististries and we want to keep the list to those we must ensure only.

PR #1111, opened 2025-07-09
update kong images

mcharriere:
Signed-off-by: Matias Charriere <matias@giantswarm.io>


marians (approved):
Feel free to also delete older tags that are no longer in use / have long been retagged.

mcharriere:
I'll do cleanup in a follow up PR, just in case.

PR #1188, opened 2026-06-04
Replace peaceiris/hugo with hugomods/hugo versions

pipo02mix:
`peaceiris/hugo` is really slow to push hugo container images, I vote for move to `hugomods/hugo`


marians (approved):
Fine, if they work for you.

Not sure why we should be retagging the older versions though.

pipo02mix:
yeah I am not sure, just in case something does not work, being able to rolled back (it has different naming)

PR #1201, opened 2026-06-25
fix: revert #1196 and restart retagging images from quay.io

hervenicol:
Revert changes from https://github.com/giantswarm/retagger/pull/1196
The removed pipeline was actually retagging images FROM quay, and we still want to do it.


mproffitt (approved):
Approved - previous PR removed the wrong thing

PR #1215, opened 2026-08-25
feat: add security-profiles-operator images

mirza-src:


mproffitt (approved):

- One space before inline comment trips yamllint's `comments` rule (warning only, exit 0) — `images/renamed-images.yaml:965`
- New entry appended after `sig-storage/snapshot-controller`, breaking the file's otherwise alphabetical order (`security-` sorts before `sig-`) — `images/skopeo-registry-k8s-io.yaml:37`
- `v0.0.1` is an invented destination tag; documented by the comment, but a future real upstream release could make it misleading — `images/renamed-images.yaml:965`

images/renamed-images.yaml
```diff
@@ -960,3 +960,7 @@
 - image: docker.io/dicebear/api
   override_repo_name: dicebear-api
   semver: ">= 4.9.0"
+- image: quay.io/security-profiles-operator/selinuxd-el9
+  override_repo_name: selinuxd
+  tag_or_pattern: "v0.0.1" # destination tag (upstream has no tags, image is pinned by sha)
```
stone-z: Personally, I'd not invent the tag and just use the SHA directly. It seems this repo will require you to set a version anyway, so maybe set it to something that will obviously never happen upstream `v0.0.0-<sha>` or something

stone-z (approved):
One suggestion, but ok with me

PR #1216, opened 2026-08-31
fix: mirror golang-adk from ghcr.io — upstream left cr.kagent.dev after 0.9.9

teemow:
Upstream stopped publishing `golang-adk` to `cr.kagent.dev` after 0.9.9 (the other kagent images still live there), so this mirror silently stalled: gsoci has golang-adk up to 0.9.9 only, while kagent 0.9.10+ controllers compose `gsoci.azurecr.io/giantswarm/golang-adk:<controller version>` for declarative go agents — every such agent pod fails with ErrImagePull (hit live on kagent 0.9.12: `golang-adk:0.9.12 not found`).

`ghcr.io/kagent-dev/kagent/golang-adk` carries all tags including the `-full` variants (verified 0.9.9, 0.9.12 and 0.9.12-full exist there).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mproffitt (approved):

PR #1219, opened 2026-09-06
fix: mirror the kagent-app -full tags — kagent 0.10.0 resolves Python agents with skills to kagent-app:<version>-full

teemow:
kagent 0.10.0 changed how the controller picks the image for a declarative agent: `resolvePythonRuntimeImage(registry, full, pinDigest)` now appends `-full` when the agent needs the SRT settings (`needsSRTSettings` = `spec.skills != nil` or `executeCodeBlocks`), the same rule the Go runtime already had (`golang-adk:<version>-full`). 0.9.9 had no Python `-full` variant at all.

Upstream publishes `app:0.10.0-full` on `cr.kagent.dev/kagent-dev/kagent/app` (and ghcr.io), but the existing `kagent-app` entry uses `semver: ">= v0.9.4"`, which excludes suffixed tags — gsoci has `kagent-app:0.9.4 … 0.10.0` and no `-full` at all. A Python agent with skills on a 0.10.0 controller therefore composes `gsoci.azurecr.io/giantswarm/kagent-app:0.10.0-full` and goes ImagePullBackOff; the Agent never reaches Ready.

This adds a second `kagent-app` entry with `tag_or_pattern: "^[0-9]+\.[0-9]+\.[0-9]+-full$"`, mirroring the pattern the `golang-adk` mirror already uses for its `-full` tags. Same source registry as the plain entry (`cr.kagent.dev` carries `0.10.0-full`; rc/beta `-full` tags are excluded by the pattern).

Verified:
- `crane ls cr.kagent.dev/kagent-dev/kagent/app | grep full` → `0.10.0-full` (plus rc/beta variants)
- `crane digest gsoci.azurecr.io/giantswarm/kagent-app:0.10.0-full` → not found today
- YAML parses (9 entries), `go vet ./...` clean

After the main run: `crane digest gsoci.azurecr.io/giantswarm/kagent-app:0.10.0-full` should match `sha256:86e0141f125b99d9504dfc280dd7f823abaa86ccc8a5b221bac2b58abf994880` (upstream).


mproffitt (approved):

- Pattern won't match a `v`-prefixed `-full` tag; upstream does occasionally publish those (`v0.10.0-rc4-full` exists today), so a future `v0.11.0-full` would be silently skipped by both this entry and the `semver: ">= v0.9.4"` one — images/renamed-kagent.yaml:15

teemow:
Coordination from the graveler kagent-outage fix (klaus-lab): graveler's kagent controller is healthy again since 2026-09-06T08:18Z — CNPG `kagent/kagent-pg` 3/3 healthy (primary kagent-pg-3, timeline 15, data intact), HelmReleases `flux-giantswarm/kagent` (0.1.47) and `agent-platform-connectivity` (3.7.2) both Ready, all three Agents Ready. The live-only `spec.values.kagent.tag: 0.9.9` on graveler was left untouched; it is safe to remove it on graveler now.

PR #1221, opened 2026-09-10
chore(agentgateway): mirror the release tags of the Giant Swarm line (giantswarm/agentgateway-upstream) into gsoci

teemow:
The Agent Platform's agentgateway now comes from a Bumblebee-owned line, the fork [giantswarm/agentgateway-upstream](https://github.com/giantswarm/agentgateway-upstream) (upstream release tag as the pin plus carried patches, built, scanned and published there — see its `FORK.md`; tracking giantswarm/giantswarm#37758). The fleet keeps pulling from gsoci, so the retagger mirrors the line's **release** tags — `vX.Y.Z-gs.N`, selected by `filter`, dev builds excluded — from `ghcr.io/giantswarm/agentgateway-upstream/{agentgateway,controller}` into the same gsoci repositories upstream's tags land in (`giantswarm/agentgateway`, `giantswarm/agentgateway-controller`); the packaging chart `giantswarm/agentgateway` and the agent-platform connectivity chart pin them from there.

The semver constraint has a `-0` pre-release floor because Masterminds/semver excludes pre-releases otherwise; the `filter` keeps only the `-gs.` tags.


mproffitt (approved):

- The source images aren't published yet: `giantswarm/agentgateway-upstream` has no git tags and no releases, and `ghcr.io/giantswarm/agentgateway-upstream/agentgateway` is `DENIED` at ghcr's anon token endpoint (a control package returns 200). `retag-renamed-images` does not `skopeo login ghcr.io` (`.circleci/config.yml:264-277`), and `listTags` only swallows `name unknown` / `repository name not known to registry` (`main.go:346-351`) — a `denied`/`unauthorized` falls through to the error path, so the daily job for this file will exit non-zero (`main.go:501`) until the fork publishes those packages publicly. Other images in the file still copy (errors are aggregated, `Fatalf` only at the end), so this is red-CI noise, not lost mirroring. images/renamed-agentgateway.yaml:20
- The PR description points at the fork's `FORK.md`; that file doesn't exist on the fork's default branch (`giantswarm`) yet, and neither does the comment's reference at images/renamed-agentgateway.yaml:15
- `>= 1.5.1-0` is unbounded across majors, so a future `v2.0.0-gs.N` would be mirrored automatically. Consistent with the existing `>= v1.2.0` entries above, so probably intended. images/renamed-agentgateway.yaml:23

PR #1224, opened 2026-09-14
fix: mirror grafana distroless image tags

hervenicol:
## What

Extend the `grafana/grafana` tag regex so we also mirror the `X.Y.Z-distroless` variants:

```diff
-    # grafana security images for grafana 10+ - tag looks like `12.0.0-security-01`
-    grafana/grafana: ^[1-9][0-9]\.[0-9]+\.[0-9]+-security-[0-9]+$
+    # grafana security images for grafana 10+ - tag looks like `12.0.0-security-01`,
+    # and the distroless variant the grafana Helm chart >= v13 defaults to - `13.2.1-distroless`
+    grafana/grafana: ^[1-9][0-9]\.[0-9]+\.[0-9]+(-security-[0-9]+(-distroless)?|-distroless)$
```

## Why

The Grafana Helm chart changed its default image tag in v13:

```yaml
# charts/grafana/values.yaml (>= 13.x)
tag: "{{ .Chart.AppVersion }}-distroless"
```

`kube-prometheus-stack` picked that subchart up in chart v89, so [kube-prometheus-stack-app](https://github.com/giantswarm/kube-prometheus-stack-app) now renders `gsoci.azurecr.io/giantswarm/grafana:13.2.1-distroless`, which we do not mirror:

```
gsoci.azurecr.io/giantswarm/grafana:13.2.1            -> OK
gsoci.azurecr.io/giantswarm/grafana:13.2.1-distroless -> MISSING
docker.io/grafana/grafana:13.2.1-distroless           -> OK
```

The `grafana/grafana: ">= 11.0.0"` entry under `images:` is a semver constraint, which does not match suffixed tags — hence extending the existing `images-by-tag-regex` entry rather than adding a second key (YAML would reject the duplicate).

This has had the `kube-prometheus-stack-app` e2e job red since its chart v90 bump: `helm install --wait` times out because the Grafana pod cannot pull its image. See giantswarm/kube-prometheus-stack-app#589.

## Why `-security-NN-distroless` is in there too

That shape is **not published today**. Checked the 1000 most recently updated `grafana/grafana` tags (of 1957 total) — the variants per release `X.Y.Z` are:

| suffix | count |
|---|---|
| *(none)* | 187 |
| `-ubuntu` | 188 |
| `-security-NN` | 54 |
| `-security-NN-ubuntu` | 54 |
| `-slim` / `-ubuntu-slim` / `-distroless` / `-distroless-slim` | 15 each |

No tag combines the two, and both plausible orderings 404 (`13.0.1-security-01-distroless`, `13.0.1-distroless-security-01`).

It is matched anyway because the combination looks likely to appear: the grafana chart has shipped security appVersions before (`12.0.0-security-01`, `12.0.1-security-01`), and since v13 its tag is a plain `tpl` of `"{{ .Chart.AppVersion }}-distroless"` with no special-casing. The branch follows the existing `-security-NN-ubuntu` precedent, with the variant suffix last.

Note this is mirroring insurance only. If grafana ships a 13.x security release *without* a distroless build, the chart renders a tag that does not exist on Docker Hub either — an upstream chart bug we cannot fix here.

## Verification

Ran the regex against those 1000 tags:

- 15 `X.Y.Z-distroless` tags now match, including the `13.2.1-distroless` we need.
- Every tag matched by the previous regex is still matched — no regression on the `-security-NN` images.
- `-distroless-slim`, floating tags like `13.0-distroless`, `-ubuntu`, single-digit majors and plain `X.Y.Z` (already covered by the semver entry) are all correctly excluded.

`images/skopeo-docker-io.yaml` still parses as valid YAML.

On the regex shape: `-distroless` appears twice. Collapsing it to `(-security-[0-9]+)?(-distroless)?` would make both groups optional and so match a bare `13.2.1`, and RE2 has no lookahead to require at least one suffix — this is the shortest form that keeps plain tags out.


hervenicol:
### Follow-up: what do distroless *security* tags look like?

Checked the 1000 most recently updated `grafana/grafana` tags (of 1957 total). The variant taxonomy per release `X.Y.Z` is:

| suffix | count |
|---|---|
| *(none)* | 187 |
| `-ubuntu` | 188 |
| `-security-NN` | 54 |
| `-security-NN-ubuntu` | 54 |
| `-slim` / `-ubuntu-slim` / `-distroless` / `-distroless-slim` | 15 each |

**No tag combines `security` and `distroless`.** Both plausible orderings 404:

```
13.0.1-security-01            -> OK
13.0.1-security-01-ubuntu     -> OK
13.0.1-security-01-distroless -> MISSING
13.0.1-distroless-security-01 -> MISSING
```

So the first version of this PR already covered everything that exists. I have extended it anyway, because the combination looks likely to appear:

- the grafana chart has shipped security appVersions before — `12.0.0-security-01` and `12.0.1-security-01`;
- since v13 its tag template is `"{{ .Chart.AppVersion }}-distroless"`, rendered by a plain `tpl` with no special-casing, so a security appVersion would render `13.x.y-security-NN-distroless`.

The added alternative follows the existing `-security-NN-ubuntu` precedent (variant suffix last). Verified against the same 1000 tags: **69 matches before, 69 after — no real tag gained or lost**, it only pre-matches the shape. `13.2.1-distroless-slim`, `13.0-distroless` and plain `13.2.1` (already covered by the semver entry) all stay excluded.

Note this is mirroring insurance only. If grafana ships a 13.x security release *without* a distroless build, the chart would render a tag that does not exist on Docker Hub either — that is an upstream chart bug we cannot fix here.

hervenicol:
Factorized the regex — nesting the optional `-distroless` inside the security branch removes the repetition:

```diff
-    grafana/grafana: ^[1-9][0-9]\.[0-9]+\.[0-9]+(-security-[0-9]+|-distroless|-security-[0-9]+-distroless)$
+    grafana/grafana: ^[1-9][0-9]\.[0-9]+\.[0-9]+(-security-[0-9]+(-distroless)?|-distroless)$
```

Match set is unchanged — diffed both against the 1000-tag corpus plus edge-case probes (`13.2.1`, `13.2.1-distroless-slim`, `13.0-distroless`, `13.2.1-ubuntu`, `9.5.1-distroless`, empty string): **zero behavioural differences**, still 69 real tags.

`-distroless` still appears twice. Collapsing it to `(-security-[0-9]+)?(-distroless)?` would make both groups optional and so match a bare `13.2.1`, which the `grafana/grafana: ">= 11.0.0"` semver entry already covers — and RE2 has no lookahead to require at least one suffix. This is the shortest form that keeps plain tags out.

TheoBrigitte (approved):

Issue #1227, opened 2026-09-18
mirror the KServe serving images and sign every mirrored digest


teemow:
## Problem

The KServe/llm-d serving stack the agent platform ships needs third-party images retagger did not mirror (`kserve/agent`, `kserve/router`, `kserve/storage-initializer`, `kserve/art-explainer`, `kserve/kserve-localmodel-controller`, `kserve/kserve-localmodelnode-agent`, `library/postgres`, `rustfs/rustfs`, `amazon/aws-cli`, the KServe controller's `quay.io/brancz/kube-rbac-proxy` sidecar), and mirrors carried no signature, so a Kyverno `verifyImages` policy that trusts one Giant Swarm identity could not admit them.

## Solution

Shipped in #1230, with #1232, #1233 and #1234 from the first sign-all passes.

- Entries in `images/skopeo-docker-io.yaml` and `images/skopeo-quay-io.yaml`: the six KServe images at `>= v0.20.0`, `postgres` `^18\.[0-9]+-alpine$`, `rustfs/rustfs` `^1\.[0-9]+\.[0-9]+(-(beta|rc)\.[0-9]+)?$`, `amazon/aws-cli` `2.17.0` (upstream tags every workday, no range), `brancz/kube-rbac-proxy` `>= v0.22.1`. Multi-arch copied as upstream ships it, landing at `gsoci.azurecr.io/giantswarm/<basename>`. `library/alpine` was already covered by the existing tag regex; `seldonio/mlserver` is not mirrored because the classic runtimes left the charts (giantswarm/kserve#81).
- Every tag copied to gsoci is signed right after the copy by `retagger sign` (cosign v3.1.3 keyless, signature stored as an OCI referrer, Rekor entry), idempotent per digest: a digest that already verifies against the job's identity is skipped. The pipeline parameter `sign-all` runs the same pass over every governed tag; it ran over the tags mirrored before this change.
- The signatures carry issuer `https://oidc.circleci.com` and subject `https://circleci.com/api/v2/projects/0a65bbac-fae8-4aca-bc8c-23d51903006b/pipeline-definitions/b8ddcac6-0699-5c43-a349-7906ac0d4a36`, the shape the architect orb signs built images with, so one attestor (issuer `https://oidc.circleci.com`, subject `^https://circleci\.com/api/v2/projects/[a-f0-9-]+/pipeline-definitions/[a-f0-9-]+$`) admits mirrored and built images alike.
[...]

PR #1232, opened 2026-09-18
fix(sign): retry the registry's throttling and the read-after-write verify

teemow:
Follow-up to #1230, found by the first `sign-all` pass (pipeline 6394).

## What

Three of the eleven gsoci `retag-registry` jobs of the pass failed in "Sign copied images": gsoci answered the burst of referrers lookups (eleven jobs × four workers) with `GET …/referrers/<digest>: TOOMANYREQUESTS`, a token the transient pattern did not know (it knew the phrase "Too Many Requests"), so those signs failed without a retry; a few verifies right after a successful sign saw `no signatures found` before the registry's referrers index caught up and were reported as "signed but does not verify".

- `TOOMANYREQUESTS` joins the transient pattern; attempts go from four to six (backoff 5, 10, 20, 40, 80 s), long enough for per-minute throttling to pass.
- The verify that follows a sign treats `no signatures found` / `no matching signatures` as a read-after-write race and retries it; the verify before a sign keeps it as the final answer for an unsigned digest.
- No change to the idempotency: a re-run of `sign-all` skips every digest that already verifies and signs the rest.

## Verified

- `go test ./...` (new test: the pre-sign verify calls cosign once on "no signatures found", the post-sign verify retries it to success; the ACR throttling line is classified transient), `go vet`, `golangci-lint` clean.
- After the merge: a second `sign-all` pipeline; its jobs end with `0 failed`, evidence on #1227.


mproffitt (approved):

- `noSignaturePattern` also matches cosign's identity-mismatch message ("no matching signatures: none of the expected identities matched"), so a genuinely mis-issued certificate burns all six attempts (~155 s) post-sign before failing — correct outcome, wasted wall-clock; `no signatures found` alone would be tighter. sign.go:74
- That interaction isn't covered: sign_test.go:89 asserts the identity-mismatch line is *not* transient, but says nothing about it now being retried when `justSigned` is true. sign_test.go:89

PR #1233, opened 2026-09-18
fix(sign): resolve the digest the registry knows, not a hash of the raw manifest

teemow:
Follow-up to #1230 and #1232, found by the first `sign-all` pass (pipeline 6394).

## What

The quay.io job of the pass failed on old mirrors — `etcd:v3.3…`, `flannel:v0.11.0-amd64`, `haproxy-exporter:v0.9.0`, `k8s-api-healthz`, `k8s-setup-network-environment`, `cilium-etcd-operator:v2.0.7` — with

```
signing digest: HEAD https://gsoci.azurecr.io/v2/giantswarm/etcd/manifests/sha256:…: unexpected status code 404 Not Found
```

Those are Docker schema 1 manifests. Their registry digest covers the payload without the `signatures` block, so the sha256 of `skopeo inspect --raw` (what `retagger sign` used) is a digest the registry does not know. For `etcd:v3.3` the raw hash is `sha256:a7057be2…` while the registry's is `sha256:4e6968ba…`.

`manifestDigest` now takes `skopeo inspect --no-tags --format '{{.Digest}}'`: the registry's digest for schema 1, schema 2 and OCI manifests, and the index digest for a multi-architecture image (checked live: `storage-initializer:v0.20.0` gives the same `sha256:9bea2bfe…` as before, `etcd:v3.3` now gives `sha256:4e6968ba…`, which `cosign triangulate` agrees with).

## Verified

- `go test ./...` (new test: the command takes skopeo's reported digest verbatim), `go vet`, `golangci-lint` clean.
- The exact skopeo call against a schema-1 and a multi-arch tag on gsoci, above.
- After the merge: the next `sign-all` pass signs the schema-1 tags; evidence on #1227.


mproffitt (approved):

- Non-`--raw` `skopeo inspect` builds the full inspect struct, so it also fetches the image config blob — one extra registry GET per tag on the same path #1232 had to add throttling retries for; `--no-tags` offsets some of it, but worth watching in the next `sign-all` pass (sign.go:148)

Issue #1238, opened 2026-09-19
retire the ghcr mirrors of the Giant Swarm lines


teemow:
## Problem

`images/renamed-kagent.yaml`, `images/renamed-substrate.yaml` and `images/renamed-agentgateway.yaml` copied Giant Swarm's own builds from `ghcr.io/giantswarm/*` into gsoci. They existed because three fork lines published to ghcr from GitHub Actions; that was a mistake, not a source to mirror. retagger is for third-party images. With the lines publishing to gsoci from CircleCI (signed with the org's identity), the copies would either race the native pushes on the same repositories or fail on a source that no longer exists.

## Solution

Every entry with a `ghcr.io/giantswarm/*` source is removed: #1242 (merge commit 23a5b644). The upstream `cr.kagent.dev/*`, `ghcr.io/kagent-dev/*` and `cr.agentgateway.dev/*` copies and the nested `fluxcd/*` controllers stay; `renamed-substrate.yaml` is gone together with its matrix row. Nothing in retagger writes to `giantswarm/kagent/*`, `giantswarm/substrate/*` or `giantswarm/agentgateway-upstream/*` any more; the flattened `giantswarm/agentgateway` and `giantswarm/agentgateway-controller` repositories carry upstream's releases only.

## Acceptance criteria

- [x] No entry names a `ghcr.io/giantswarm/*` source — #1242.
- [x] The nested `fluxcd/*` entries and the upstream copies stay and the sweep is green — the entries stay (#1242); the first `main` run after the merge, [pipeline 6426](https://app.circleci.com/pipelines/github/giantswarm/retagger/6426/workflows/d39de6a8-e4d9-4893-912d-54a70af4c403), finished green (validation 3/3, every `retag-renamed-images` job success).
- [x] The repositories the lines now push to carry only natively built, signed tags from that point on — no retagger entry targets them (#1242).

## Related

#1229 (the mirrors this retires), epic giantswarm/giantswarm#37853.

teemow:
Removed in #1242, merged as 23a5b644 (squash). The PR's checks were green after a title fix for the conventional-commit check; `git grep ghcr.io/giantswarm` finds nothing on `main`.

[...]

PR #1242, opened 2026-09-19
chore(images): retire the ghcr.io/giantswarm mirrors of the fork lines

teemow:
## What

retagger mirrors third-party images only. The three fork lines (kagent, Substrate, agentgateway) publish to gsoci from their own pipelines now, so their `ghcr.io/giantswarm/*` sources are no longer something to copy — a copy would either race the native pushes on the same repositories or fail on a source that stops existing.

- `images/renamed-kagent.yaml`: the four `ghcr.io/giantswarm/kagent/*` entries and their comment block removed. The upstream `cr.kagent.dev/*` and `ghcr.io/kagent-dev/*` entries stay.
- `images/renamed-agentgateway.yaml`: the two `ghcr.io/giantswarm/agentgateway-upstream/*` entries and their comment block removed. The `cr.agentgateway.dev/*` copies into `agentgateway` and `agentgateway-controller` stay; a one-line comment states that these two repositories carry upstream's releases while the Giant Swarm line publishes under `gsoci.azurecr.io/giantswarm/agentgateway-upstream/`.
- `images/renamed-substrate.yaml`: deleted — it held nothing but `ghcr.io/giantswarm/substrate/*` entries — and its row removed from the `retag-renamed-images` matrix in `.circleci/config.yml`.

No entry copies a Giant Swarm build any more; from this change on the repositories the lines push to carry only natively built, signed tags. The nested `fluxcd/*` controllers in `renamed-images.yaml` are untouched.

## Verification

- `git grep ghcr.io/giantswarm` finds nothing.
- `yamllint --config-file .yamllint images/*.yaml` passes locally (the same check `validate-images-yaml` runs).
- `retag-renamed-images` only runs on `main`; its matrix names exactly the files that exist.

Refs #1238


mproffitt (approved):

- The remaining comment says the GS agentgateway line publishes under `gsoci.azurecr.io/giantswarm/agentgateway-upstream/`, but the removed entries landed those `-gs.N` tags in `giantswarm/agentgateway` / `giantswarm/agentgateway-controller` — consumers pinning the old flat path need updating in their own repos (images/renamed-agentgateway.yaml:1)

